PR #161, opened 2026-04-25
feat(applications): expose destination_uuid in create_* tool schema

barbe1bc:
## Summary

The Coolify v4 API accepts an optional `destination_uuid` on every application-create endpoint (`/applications/public`, `/applications/private-github-app`, `/applications/private-deploy-key`, `/applications/dockerimage`). The TypeScript request types in `src/types/coolify.ts` already declare `destination_uuid?: string`, so the underlying client surface is ready — but the MCP tool schema in `src/lib/mcp-server.ts` omits it, so callers of the `application` tool can't specify which destination (Docker network) on a multi-destination server should host the new app. They get the server's default destination silently, which is wrong on hosts that have more than one Docker network configured in Coolify.

## Changes

- Add `destination_uuid: z.string().optional()` to the `application` tool input schema (next to the other `*_uuid` fields).
- Thread `destination_uuid: args.destination_uuid` through all four create dispatch arms: `create_public`, `create_github`, `create_key`, `create_dockerimage`.
- Add one client test asserting `destination_uuid` is passed through to the request body in `createApplicationPublic`.

No other behavior changes. The field stays optional, so existing callers are unaffected.

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 263 passed (262 existing + 1 new)
- [x] `npm run lint` — no new warnings/errors (the 4 pre-existing warnings in unrelated code are unchanged)


StuMason (approved):
Nice, thanks!

PR #162, opened 2026-04-25
fix(applications): pass domains, instant_deploy, custom_docker_run_options, custom_labels to Coolify API

barbe1bc:
> Stacked on top of #161. Rebases to main cleanly once that merges.

### 1. Bug — fqdn silently dropped on three create paths

The `mapFqdnToDomains()` translation in the client was applied on the public / private-github-app / private-deploy-key / update paths, but **not** on `dockerfile`, `dockerimage`, or `dockercompose`. So calling

```jsonc
{
  \"action\": \"create_dockerimage\",
  \"docker_registry_image_name\": \"traefik/whoami\",
  \"ports_exposes\": \"80\",
  \"fqdn\": \"https://whoami.example.com\",
  \"...\": \"...\"
}
```

sent the literal key `fqdn` to Coolify, which dropped it. The app got created with no domain and no error surfaced to the caller — pure silent failure.

### 2. No way to pass `domains` directly

Callers naturally try `domains` because the Coolify API documents that name **and** because `get_application` returns the field labeled `fqdn` in responses. Both should work. Now both do — if both are present, explicit `domains` wins.

### 3. Three useful API fields weren't exposed

Coolify already accepts these on create + update; they were just absent from the MCP surface:

| Field | Why it's useful |
|---|---|
| `instant_deploy: boolean` | Create + deploy in one call. (Was already in the TS request types but not in the MCP schema.) |
| `custom_docker_run_options: string` | Add flags like \`--network=...\` without going to the UI. |
| `custom_labels: string` (base64) | Custom Traefik routing rules beyond what FQDN auto-generates. |

## Changes

**`src/lib/coolify-client.ts`**
- Apply `mapFqdnToDomains()` to `createApplicationDockerImage`, `createApplicationDockerfile`, `createApplicationDockerCompose`.
- Update `mapFqdnToDomains()` to prefer an explicit `domains` value over the legacy `fqdn` alias when both are present.

**`src/types/coolify.ts`**
- Add `domains?: string`, `custom_docker_run_options?: string`, `custom_labels?: string` to all 6 `CreateApplication*Request` types and to `UpdateApplicationRequest`. (`instant_deploy` was already typed.)

**`src/lib/mcp-server.ts`**
- Add `domains`, `custom_docker_run_options`, `custom_labels`, `instant_deploy` to the `application` tool input schema.
- Thread all four through the four `create_*` dispatch arms. `update` picks them up via the existing rest-spread.

## Tests (`+5`, 268 total)

- `dockerimage`: `fqdn → domains` mapping (regression test for the bug)
- `dockerfile`: same
- `dockercompose`: same
- explicit `domains` takes precedence over `fqdn` when both are passed
- `instant_deploy`, `custom_docker_run_options`, `custom_labels` pass through on `dockerimage`

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 268 passed (263 existing + 5 new)
- [x] Live verification deferred: I'm wiring this up in production via a fork-build right after merge; happy to add the run output here if useful.

### DIFF
--- a/src/__tests__/coolify-client.test.ts
+++ b/src/__tests__/coolify-client.test.ts
@@ -1202,6 +1202,88 @@ describe('CoolifyClient', () => {
       expect(callBody.fqdn).toBeUndefined();
     });
 
+    it('should map fqdn to domains in createApplicationDockerImage', async () => {
+      mockFetch.mockResolvedValueOnce(mockResponse({ uuid: 'new-app-uuid' }));
+
+      await client.createApplicationDockerImage({
+        project_uuid: 'proj-uuid',
+        server_uuid: 'server-uuid',
+        docker_registry_image_name: 'traefik/whoami',
+        ports_exposes: '80',
+        fqdn: 'https://whoami.example.com',
+      });
+
+      const callBody = JSON.parse(mockFetch.mock.calls[0][1]?.body as string);
+      expect(callBody.domains).toBe('https://whoami.example.com');
+      expect(callBody.fqdn).toBeUndefined();
+    });
+
+    it('should map fqdn to domains in createApplicationDockerfile', async () => {
+      mockFetch.mockResolvedValueOnce(mockResponse({ uuid: 'new-app-uuid' }));
+
+      await client.createApplicationDockerfile({
+        project_uuid: 'proj-uuid',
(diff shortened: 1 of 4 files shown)


StuMason (approved):
Really nice, thanks!

PR #167, opened 2026-04-28
feat: add --header flag for custom HTTP headers

imantsk:
## Summary
- Adds `--header` CLI flag to inject custom HTTP headers on all API requests
- Multiple `--header` flags can be specified (e.g., `--header "CF-Access-Client-Id: value" --header "CF-Access-Client-Secret: value"`)
- Headers are applied to all outgoing requests alongside the existing `Authorization: Bearer` header
- Useful for Cloudflare Zero Trust, custom auth proxies, and other middleware that requires additional headers

## Changes
- `CoolifyConfig` interface: added optional `customHeaders` field
- `CoolifyClient`: spreads `customHeaders` into all fetch requests (both `request()` and `getVersion()`)
- `index.ts`: parses `--header "Name: Value"` from argv, passes to config

## Usage
```bash
npx @masonator/coolify-mcp \
  --header "CF-Access-Client-Id: abc123.access" \
  --header "CF-Access-Client-Secret: secret"
```

## Test plan
- [x] Build succeeds (`npm run build`)
- [x] Verified headers appear on outbound requests using echo server
- [x] MCP tools/list works with custom headers
- [x] Multiple --header flags coexist correctly

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -72,6 +72,7 @@
       "integrity": "sha512-CGOfOJqWjg2qW/Mb6zNsDm+u5vFQ8DxXfbM09z69p5Z6+mE1ikP2jUXw+j42Pf1XTYED2Rni5f95npYeuwMDQA==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@babel/code-frame": "^7.29.0",
         "@babel/generator": "^7.29.0",
@@ -1762,6 +1763,7 @@
       "integrity": "sha512-TI1XGwKbDpo9tRW8UDIXCOeLk55qe9ZFGs8MTKU6/M08HWTw52DD/IYhfQtOEhEdPhLMT26Ka/x7p70nd3dzDg==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@typescript-eslint/scope-manager": "8.59.0",
         "@typescript-eslint/types": "8.59.0",
@@ -2255,6 +2257,7 @@
       "integrity": "sha512-UVJyE9MttOsBQIDKw1skb9nAwQuR5wuGD3+82K6JgJlm/Y+KI92oNsMNGZCYdDsVtRHSak0pcV5Dno5+4jh9sw==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "bin": {
         "acorn": "bin/acorn"
       },
@@ -2603,6 +2606,7 @@
         }
       ],
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "baseline-browser-mapping": "^2.10.12",
         "caniuse-lite": "^1.0.30001782",
@@ -3336,6 +3340,7 @@
       "integrity": "sha512-wiyGaKsDgqXvF40P8mDwiUp/KQjE1FdrIEJsM8PZ3XCiniTMXS3OHWWUe5FI5agoCnr8x4xPrTDZuxsBlNHl+Q==",
       "dev": true,
       "license": "MIT",
+      "peer": true,
       "dependencies": {
         "@eslint-community/eslint-utils": "^4.8.0",
         "@eslint-community/regexpp": "^4.12.2",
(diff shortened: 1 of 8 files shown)


StuMason:
Hey @imantsk — thanks, this is a solid use case (CF Zero Trust + auth proxies in front of Coolify is real and underserved), and the core design is small and clean. I want to land it. A few things to sort first though.

## Blocker: `package.json` changes will break consumers

Two changes there I don't think were intentional:

```diff
-    "prepare": "husky",
+    "prepare": "husky || true",
+    "postinstall": "npm run build || true",
```

The `postinstall: npm run build` is the big one. The package ships with pre-built `dist/` already — adding `postinstall` means **every user running `npx @masonator/coolify-mcp`** triggers a TypeScript compile on their machine. That:

- Slows cold start dramatically (~10s+ tsc invocation)
- Fails on minimal environments where dev deps aren't installed
- Silently swallows the failure with `|| true`, so users see broken installs with no log

The `husky || true` change has the same swallow-the-error problem and isn't needed for the feature.

**Could you revert both `package.json` lines?** They look like local dev-environment workarounds rather than intentional package changes. The feature itself doesn't need them.

## Should-have: tests

CLAUDE.md is explicit about test coverage on new client methods:

> codecov will fail PRs with uncovered lines. Always run `npm test` before committing.

Codecov passed here on the surface (the new lines have low cyclomatic count) but the project convention is "new public surface gets tests." A couple of small ones would close it out:

- `parseHeaders` unit test in a new test file (or add a section to an existing one) covering: single header, multiple headers, malformed (no colon → ignored), trim behaviour
- `CoolifyClient` test verifying `customHeaders` get merged into outgoing fetch headers (mock fetch, assert call args)

The pattern is well-established in `src/__tests__/coolify-client.test.ts` — should be 30 lines total.

## Optional: header-precedence note

Spread order in `request()` puts `customHeaders` AFTER the bearer `Authorization`:

```ts
headers: {
  'Content-Type': 'application/json',
  Authorization: `Bearer ${this.accessToken}`,
  ...this.customHeaders,        // wins if user passes Authorization
  ...options.headers,
}
```

Means a user passing `--header "Authorization: ..."` would silently override the Coolify token — not what they want for "additive proxy headers." Two reasonable answers:

1. Filter `customHeaders` to drop reserved names (`Authorization`, `Content-Type`) with a console warning
2. Just document the override behaviour and leave it

I'd lean (1), but (2) is acceptable. Up to you.

## Path to merge

If you can:

1. Revert the `package.json` changes
2. Add the two unit tests above
3. (Optional) Decide on header-precedence behaviour and document/code

…I'm happy to merge and ship it in v2.8.1 alongside the env-var masking and `/version` parse fixes. Genuine thanks for the contribution — the diff itself is exactly the right shape, just needs those package.json reverts before we can let it through.

imantsk:
Thanks for the quick and thorough review, @StuMason! Apologies for the unintended `package.json` changes, those were local dev workarounds to enable running the MCP server via `npx` from my fork, didn't mean to include them in the PR.

Addressed all three points:
1. **Reverted `package.json`** — both `prepare` and `postinstall` lines restored/removed
2. **Added tests** — `parseHeaders` unit tests (7 cases) + `customHeaders` merge verification + reserved header filtering test
3. **Header precedence** — went with option (1): constructor filters `Authorization` and `Content-Type` from custom headers with a `console.warn`

Extracted `parseHeaders` to its own module since `index.ts` runs `main()` on import, making direct import in tests impractical.

StuMason:
This is excellent work, @imantsk — thank you for turning it around so quickly.

Re-read the diff:

- **package.json reverts** ✅ — clean
- **`parseHeaders` extracted to its own module** — good call, the `main()`-on-import constraint was real, and the dedicated module makes this much more testable
- **Tests** — genuinely thorough. 7 cases on `parseHeaders` covering edge cases I wouldn't have asked for (value with colons, missing value after `--header`, whitespace trim). The reserved-header filter test asserts both behaviour (filtered/not-filtered headers) and the warning fires per ignored entry. Better than what I'd have written.
- **Reserved-header filter** — case-insensitive on the lookup, warns per entry, applied at constructor time so the cost is paid once. Clean.

CI is fully green. I'm happy to merge.

The only two follow-ups I'll handle at release time, no need for you to touch:

1. README — add the `--header` flag example to the install/config section
2. CHANGELOG `[Unreleased]` entry crediting you

This bundles into v2.8.1 alongside the env-var masking and `/version` parse fixes. Will mention you in the release notes. Genuine thanks for the contribution and the responsiveness — top-tier PR experience.

StuMason (approved):
Nice, love this!

PR #172, opened 2026-04-30
feat: add 7 new tools and expand 3 existing (38 → 45 total)

opastorello:
## Summary

- Adds **7 new MCP tools**: `storages`, `scheduled_tasks`, `delete_preview`, `hetzner`, `list_resources`, `health`, `api_control`
- Expands **3 existing tools**: `env_vars` (database support + bulk_update), `github_apps` (list_repos + list_branches), `database_backups` (delete_execution)
- Covers **42 previously missing Coolify API endpoints**, identified by comparing the OpenAPI spec from the v4.x branch against the existing implementation

## Changes

**`src/types/coolify.ts`** — New types for storages, scheduled tasks, Hetzner cloud, and GitHub App repositories

**`src/lib/coolify-client.ts`** — 40 new client methods:
- Application/database/service persistent storage CRUD
- Application/service scheduled task CRUD + execution listing
- Application preview deletion
- Database env var CRUD + bulk update
- Backup execution deletion
- Service env var bulk update
- Hetzner cloud (locations, server types, images, SSH keys, create server)
- GitHub App repo/branch listing
- Resource listing, health check, API enable/disable

**`src/lib/mcp-server.ts`** — 7 new tools + 3 expanded tools following the project's consolidated action-parameter pattern to minimize token usage

**`README.md`** / **`CLAUDE.md`** — Updated tool count (38 → 45) and tool tables

## Tests

- **42 new test cases** (279 → 321 total)
- All client methods covered in `coolify-client.test.ts` (HTTP method, URL, body, return value)
- All methods verified in `mcp-server.test.ts` (existence checks)
- Coverage: **97.98% statements**, 84.61% branches, 99.52% functions, 98.84% lines

## Test plan

- [x] `npm run build` — clean
- [x] `npm test` — 321 passed, 0 failed
- [x] `npm run lint` — clean (pre-commit hooks passed)
- [ ] Live smoke test against Coolify instance (deferred — no access to live server in this environment)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,16 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Changed
+
+- **`delete_preview` moved into `application` tool** — `delete_preview` is now `action: 'delete_preview'` on the existing `application` tool (alongside create/update/delete), consistent with the v2.0.0 consolidation pattern. The standalone `delete_preview` tool is removed.
+- **`system` tool consolidates health/resource/API controls** — `health`, `list_resources`, and `api_control` are removed as standalone tools and replaced by a single `system` tool with `action: 'health' | 'list_resources' | 'enable_api' | 'disable_api'`. Net tool count: 45 → 42.
+- **`storages` `update` action now requires `storage_uuid`** — Previously `storage_uuid` was not validated for `update`, causing the PATCH to send without the required field. Also fixed data construction to occur inside each action branch (after validation), so non-null assertions are guaranteed.
+
+### Fixed
+
+- **`listResources()` explicit return type** — Changed from `Promise<unknown>` to `Promise<ResourceListItem[]>` with a new `ResourceListItem` interface in `src/types/coolify.ts`, complying with the project's no-implicit-any policy.
+
 ## [2.10.0] - 2026-05-14
 
 ### Added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -144,6 +144,6 @@ When making changes to the codebase, ensure documentation is updated:
    - New example prompts needed
    - Response size improvements made (update comparison table)
 
-3. **This file (CLAUDE.md)** - Update tool count if changed (currently 38 tools)
+3. **This file (CLAUDE.md)** - Update tool count if changed (currently 45 tools)
 
 Always work on a feature branch and include documentation updates in the same PR as code changes.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,7 +4,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 ## Project Overview
 
-MCP (Model Context Protocol) server for Coolify that provides 38 token-optimized tools for AI assistants to manage infrastructure through natural language. Tools cover servers, projects, environments, applications, databases, services, deployments, private keys, teams, cloud tokens, documentation search, smart diagnostics, and batch operations. v2.0.0 reduced token usage by 85% (from ~43,000 to ~6,600 tokens) by consolidating related operations into single tools with action parameters.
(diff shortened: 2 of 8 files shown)


StuMason (changes requested):
Thanks @opastorello — this is high-quality work and most of it should land. A few changes needed before merge.

## What I love

The five consolidated tools (`storages`, `scheduled_tasks`, `hetzner`) and three expansions (`env_vars`, `github_apps`, `database_backups`) all follow the action-parameter pattern from v2.0.0 — exactly right. Types are explicit, file structure is correct, codecov passes. The `claude-review` workflow failure is an infra issue on my side (Actor permissions on fork PRs), not a real signal.

## Required changes

**1. Token budget — fold three single-purpose tools into existing ones.** v2.0.0's whole point was reducing tool count to cut MCP context size by 85%. Three of the new top-level tools are exactly the pattern we collapsed:

- `delete_preview` → add as a `delete_preview` action on the existing `application` tool (where get/restart/stop/start already live).
- `health` + `list_resources` + `api_control` → collapse into one tool (call it `system` or extend `get_infrastructure_overview`) with `action: 'health' | 'list_resources' | 'enable_api' | 'disable_api'`.

That brings the count from 38 → 41 (net +3) instead of 38 → 45 (net +7), which I'm much happier signing off on. If you can show the token-budget delta is negligible per tool I'd reconsider, but as a default policy a 7-tool addition needs explicit per-tool justification.

**2. `listResources(): Promise<unknown>`** at `src/lib/coolify-client.ts:1326`. CLAUDE.md mandates explicit return types and the project pattern is named interfaces in `src/types/coolify.ts`. The `/resources` response is enumerable (servers, apps, services, databases) — please add a `ResourceListItem` type.

**3. Hetzner endpoints aren't in our OpenAPI chunks.** They came from "the v4.x branch" and the PR notes the live smoke test was skipped. Per CLAUDE.md's "Coolify API Gotchas" section, untested endpoints carry real risk. Please run a live smoke test (`/smoke-test` or `npm run test:integration`) against a Coolify with Hetzner configured before we merge. Also: the Hetzner methods build query strings via manual concatenation; the rest of the client uses the `request()` wrapper — minor consistency nit.

**4. `storages` `update` action** at `src/lib/mcp-server.ts:1608`. The data object is built with `args.mount_path!` non-null assertion before the validity guard runs — and the `update` case only checks `type`, not `storage_uuid`, but the underlying PATCH puts `storage_uuid` in the body. Two fixes:
  - Construct `createData`/`updateData` inside each case branch *after* validation, so the `!` assertions are real.
  - Require `storage_uuid` for the `update` action.

## Non-blocking

The MCP-server test additions are all method-existence checks — that matches the established codebase pattern, so coverage passes via the client tests. Just noting that the handler routing logic (where bug surface for an expansion PR mostly lives) is therefore unexercised end-to-end. Not blocking this PR but worth tracking.

Let me know if you want to pair on the consolidations.

StuMason (approved):
Thanks @opastorello — the revision addresses everything I asked for:

- `delete_preview` folded into `application` tool ✓
- `health` + `list_resources` + `api_control` collapsed into `system` (net 38 → 42, well-justified) ✓
- `ResourceListItem` type with concrete fields, `Promise<unknown>` gone ✓
- `storages` `update` now requires `storage_uuid` ✓ (note: the methods table is still built at the top of the handler with non-null assertions, but the runtime guards run first so the arrow callbacks only execute after validation. Functionally correct — happy to leave as-is.)

## What I additionally verified

- 349/349 tests pass locally, 98.88% coverage
- Pushed a prettier fix for `mcp-server.ts` (`a582edb`) to clear format check
- **Live smoke-tested against a real Coolify instance**:
  - `system.health` action → `GET /api/v1/health` → 200 "OK"
  - `system.list_resources` action → `GET /api/v1/resources` → 200 with full resource list (340KB)
  - All 4 Hetzner endpoints (locations, server-types, images, ssh-keys) return 401 "Unauthenticated" — the routes **exist** (not 404) and are gated by an auth scope my token doesn't have, which makes sense for Hetzner provisioning. The wiring is correct; whether they work end-to-end at runtime depends on the caller having a properly-scoped token and a Hetzner cloud-provider token UUID configured. Documented the path as `/hetzner/server-types` (hyphen, matches your client code — the underscore variant returns 404).

The "no live access" caveat in the PR description is fine — the smoke test is now done.

Merging. Will roll into the next release.

PR #174, opened 2026-05-01
fix(env-vars): match Coolify API field names (is_buildtime/is_runtime), expose runtime-only flag

kashik0i:
## Summary

Fixes #135. Coolify v4 API uses `is_buildtime` and `is_runtime` (one word each), not `is_build_time` (with underscore). The MCP was using the wrong field name throughout, which produced two distinct silent-failure modes against a real Coolify server.

## Live-API verification (Coolify v4.0.0-beta.473)

```
# POST /applications/{uuid}/envs   with is_build_time -> 422
{"message":"Validation failed.","errors":{"is_build_time":["This field is not allowed."]}}

# PATCH /applications/{uuid}/envs  with is_build_time -> 422 (same)

# PATCH /applications/{uuid}/envs/bulk with is_build_time -> 201
# but the flag is silently dropped (response: "is_buildtime": true unchanged)

# All three endpoints with is_buildtime/is_runtime -> 201 + flag persists.
```

## What changes

| Surface | Before | After |
|---|---|---|
| `EnvironmentVariable`, `CreateEnvVarRequest`, `UpdateEnvVarRequest`, `EnvVarSummary`, diagnostics env-var view | `is_build_time` (always `undefined` against real Coolify) | `is_buildtime` + `is_runtime` |
| `env_vars` tool — app create/update, service create | flags not exposed; tool dropped them with a TODO comment | accepts `is_buildtime` + `is_runtime`, threads to client |
| `bulk_env_update` tool / `bulkEnvUpdate(...)` | `is_build_time` was sent on the wire and silently ignored | `is_buildtime` + new `is_runtime` |
| `toEnvVarSummary` and diagnostics projection | read `envVar.is_build_time` (always `undefined`) | read the correct fields |

## Why this matters (the #135 use case)

Setting `is_buildtime: false, is_runtime: true` is the only way to add multiline secrets (PEM keys, certificates, etc.) to a Coolify app via API/MCP. Without it, Coolify injects the value as a Dockerfile `ARG` and breaks the build with a parser error on the second line of the PEM. We've hit this same trap multiple times in production (NODE_ENV, Auth.js `AUTH_URL`, `PASSPORT_PRIVATE_KEY`-style secrets) and worked around it via the Coolify web UI.

## Breaking change (typed-only)

- `EnvironmentVariable.is_build_time` -> `is_buildtime` + new `is_runtime`
- `Create/UpdateEnvVarRequest.is_build_time?` -> `is_buildtime?` + new `is_runtime?`
- `EnvVarSummary.is_build_time` -> `is_buildtime` + new `is_runtime`
- `bulkEnvUpdate(uuids, key, value, isBuildTime)` -> `bulkEnvUpdate(uuids, key, value, isBuildtime?, isRuntime?)`

Programmatic TypeScript consumers of `@masonator/coolify-mcp` will need to update field references. Note: under the old field name the value was always `undefined` against a real Coolify server, so any caller that worked end-to-end was already not relying on it.

## Notes

- Supersedes the `claude/issue-135-20260303-0903` branch (which used the wrong field name `is_build_time` and would have produced HTTP 422 on every create/update call). The deleted comment `// is_build_time is not passed - Coolify API rejects it for create action` was correct in spirit — the workaround was just hiding the underlying field-name bug.
- Service `update` action remains hard-erroring (`Error: service env update not supported`); preserved as-is, separate from this fix.

## Test plan

- [x] `npm run build` clean.
- [x] `npm test` — 282/282 passing (was 280/280, +2 new tests covering the runtime-only create + update case from #135).
- [x] `npm run lint` — 0 errors (4 pre-existing warnings unchanged).
- [x] Live-API probe against Coolify v4.0.0-beta.473 — POST/PATCH/bulk all accept `is_buildtime` + `is_runtime` and persist the flags; old field name `is_build_time` either 422s (single endpoints) or silently no-ops (bulk).

### DIFF
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -100,6 +100,7 @@ The Coolify OpenAPI docs are unreliable — always test against the real API. Kn
 
 - **`docker_compose_raw` requires base64** — The API expects base64-encoded YAML, but the field name suggests raw content. The client auto-encodes this field so models and callers can pass plain YAML.
 - **Validation errors vary in format** — The `errors` field in API error responses can contain `string[]` or plain `string` values. The client handles both.
+- **Env var field names are `is_buildtime` and `is_runtime`** (one word each), not `is_build_time` (two words). On `POST /applications/{uuid}/envs` and `PATCH /applications/{uuid}/envs` the wrong name returns HTTP 422 `"This field is not allowed."`; on `PATCH /applications/{uuid}/envs/bulk` the wrong name is silently ignored (request returns 201 but the flag stays at the default). Verified against Coolify v4.0.0-beta.473 in #174 / #135. When adding env-var related code or tests, mirror the API field names exactly — do not paraphrase to `is_build_time`.
 
 ## TypeScript Standards
 
--- a/src/__tests__/coolify-client.test.ts
+++ b/src/__tests__/coolify-client.test.ts
@@ -3622,6 +3622,22 @@ describe('CoolifyClient', () => {
           }),
         );
       });
+
+      it('should send only is_runtime when buildtime is left undefined', async () => {
+        mockFetch
+          .mockResolvedValueOnce(mockResponse(mockApps))
+          .mockResolvedValueOnce(mockResponse({ message: 'Updated' }));
+
+        await client.bulkEnvUpdate(['app-1'], 'API_KEY', 'val', undefined, false);
+
+        expect(mockFetch).toHaveBeenCalledWith(
+          'http://localhost:3000/api/v1/applications/app-1/envs',
+          expect.objectContaining({
+            method: 'PATCH',
+            body: JSON.stringify({ key: 'API_KEY', value: 'val', is_runtime: false }),
+          }),
+        );
+      });
     });
 
     describe('stopAllApps', () => {
--- a/src/__tests__/coolify-client.test.ts
+++ b/src/__tests__/coolify-client.test.ts
@@ -1612,7 +1612,8 @@ describe('CoolifyClient', () => {
       uuid: 'env-var-uuid',
       key: 'API_KEY',
       value: 'secret123',
-      is_build_time: false,
+      is_buildtime: false,
+      is_runtime: true,
     };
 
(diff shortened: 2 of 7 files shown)


kashik0i:
End-to-end smoke test through the MCP layer (not just the HTTP client) against our Coolify v4.0.0-beta.473 instance:

Built the branch, reconnected the local MCP, called the new tool surface:

```
mcp__coolify__env_vars
  action=create resource=application uuid=<app>
  key=PROBE_MCP_E2E value=probe-via-mcp
  is_buildtime=false is_runtime=true
-> { uuid: "..." }
```

Curl-verified the flags actually persisted server-side (both the main env and the auto-created preview mirror):

```
{ "key": "PROBE_MCP_E2E", "is_buildtime": false, "is_runtime": true, "is_preview": false }
{ "key": "PROBE_MCP_E2E", "is_buildtime": false, "is_runtime": true, "is_preview": true }
```

Then the update path with both flags flipped:

```
mcp__coolify__env_vars
  action=update resource=application uuid=<app>
  key=PROBE_MCP_E2E value=probe-via-mcp-updated
  is_buildtime=true is_runtime=false
-> { is_buildtime: true, is_runtime: false, ... }
```

Coolify response confirms the flip applied. Probe env vars cleaned up afterwards.

Confirms the full chain works: zod schema → handler destructure → DTO field → request body → Coolify API → flag persistence. No layer between MCP and wire silently dropping fields.

StuMason (approved):
Coverage now 100% on the patch (added a fallback-branch test for `?? false`/`?? true`). Live-tested by author against Coolify v4.0.0-beta.473, CLAUDE.md now documents the API field-name gotcha so future contributors don't get caught.

Pushed two small additions to your branch:
- `test(diagnose)` covering the legacy-response path where Coolify omits the flags entirely
- `test(env_vars)` exercising the MCP handler passthrough via the SDK's `_registeredTools` map

Thanks for the careful work and the live-server verification. Closing #135 with this.

PR #177, opened 2026-05-10
fix: parse non-json Coolify responses defensively

### LINKED ISSUE #163: JSON.parse error when /api/v1/version returns plain text (Coolify 4.0.0-beta.474)
### Description
The MCP server fails with the following error when attempting to fetch the Coolify version or perform other operations:
`SyntaxError: Unexpected non-whitespace character after JSON at position 4 (line 1 column 5)`

This occurs because some versions of Coolify (specifically `4.0.0-beta.474`) return the version number as a raw string (e.g., `4.0.0-beta.474`) from the `/api/v1/version` endpoint, while the `CoolifyClient.request` method attempts to `JSON.parse` the response text unconditionally if it is not empty.

### Steps to Reproduce
1. Connect to a Coolify instance (v4.0.0-beta.474).
2. Run any tool that initializes the client (e.g., `get_infrastructure_overview`).
3. Observe the parsing error.

### Proposed Fix
Modify `src/lib/coolify-client.ts` to check the `Content-Type` header before parsing.

```typescript
      const text = await response.text();
      const contentType = response.headers.get('Content-Type');
      let data: any = {};

      if (text) {
        if (contentType?.includes('application/json')) {
          try {
            data = JSON.parse(text);
          } catch (e) {
            data = text;
          }
        } else {
          data = text;
        }
      }
```

I have verified this fix locally, and it resolves the issue.

zxyasfas:
## Summary

Fixes #163.

## Changes

- Parse `request<T>` responses based on the response `Content-Type` instead of unconditionally calling `JSON.parse`.
- Fall back to raw text for `text/plain` responses and malformed JSON bodies while preserving JSON parsing for JSON and `+json` content types.
- Add regression coverage for plain-text application logs and malformed JSON responses.

## Test plan

- [x] Tests added/updated
- [x] Manually tested locally

Validation run locally:

- `$env:NODE_OPTIONS='--experimental-vm-modules'; npx jest --testPathIgnorePatterns=integration --runInBand`
- `$env:NODE_OPTIONS='--experimental-vm-modules'; npx jest --testPathIgnorePatterns=integration --runTestsByPath src/__tests__/coolify-client.test.ts --runInBand --coverage=false`
- `npm run build`
- `npm run lint`
- `npm run format:check`

`npm run lint` passes with the repository's existing warning-only `no-explicit-any` reports.

## Checklist

- [x] Documentation updated if needed

### DIFF
--- a/src/__tests__/coolify-client.test.ts
+++ b/src/__tests__/coolify-client.test.ts
@@ -3,12 +3,19 @@ import { CoolifyClient } from '../lib/coolify-client.js';
 import type { ServiceType, CreateServiceRequest } from '../types/coolify.js';
 
 // Helper to create mock response
-function mockResponse(data: unknown, ok = true, status = 200): Response {
+function mockResponse(
+  data: unknown,
+  ok = true,
+  status = 200,
+  contentType = 'application/json',
+): Response {
+  const body = contentType.includes('application/json') ? JSON.stringify(data) : String(data);
   return {
     ok,
     status,
     statusText: ok ? 'OK' : 'Error',
-    text: async () => JSON.stringify(data),
+    headers: new Headers({ 'Content-Type': contentType }),
+    text: async () => body,
   } as Response;
 }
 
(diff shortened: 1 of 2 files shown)


StuMason (approved):
Defensive parse is exactly what issue #163 needed. Two intentional improvements over my sketch — both good:

- Empty content-type → treated as JSON (backward-compat with mocks/servers that omit the header)
- `+json` suffix accepted (e.g. `application/problem+json`)

Both failure modes are covered by tests, the `getVersion()` direct-text carve-out is untouched, builds and codecov both green. Thanks @zxyasfas.

PR #185, opened 2026-05-13
fix(application): wire build-config and health_check fields through create_* (#178)

### LINKED ISSUE #178: application: `dockerfile_location`, healthcheck fields, and several build params not wired through create_* (and dockerfile_location missing from update schema)
Two related gaps in the `application` tool that I hit while bringing up a NestJS+Vite monorepo on Coolify.

### 1. `dockerfile_location` is not in the zod schema at all

In `src/lib/mcp-server.ts` the `application` tool's schema (around lines 458–495) declares health check fields, basic create fields, and update fields — but it does not declare `dockerfile_location` (nor `dockerfile_target_build`, `base_directory`, `publish_directory`, `install_command`, `build_command`, `start_command`, `watch_paths`). All of these are present on `Application` (get response) and on the `UpdateApplicationRequest` type in `src/types/coolify.ts` (line 365), so they're known to be valid on Coolify's API.

Effect: passing `dockerfile_location` to `update` is silently stripped by zod before the spread, so the field never reaches Coolify. There is no way to set the Dockerfile path for a monorepo via MCP today; the workaround is `curl PATCH …` with the same field.

### 2. Health check fields are accepted on `update` but ignored on every `create_*`

The `update` case (mcp-server.ts ~line 648) spreads all schema-validated args into `updateApplication`, which is why issue #62's request worked — for update. But the four `create_*` cases (`create_public`, `create_github`, `create_key`, `create_dockerimage`) hand-pick fields one by one, and none of them includes any `health_check_*` field, even though those fields are declared in the same tool's schema.

Effect: `create_key`/`create_github`/etc. with `health_check_enabled: true, health_check_path: '/health', …` returns `{uuid, domains}` and a `get_application` immediately afterwards shows `health_check_enabled: false` and the default path. To get healthchecks set, the only path is to follow the `create` with a `update` round-trip (or curl PATCH).

The same hand-picking ignores any of the build-config fields listed in §1, even though `CreateApplicationPublicRequest` (types/coolify.ts:250) includes `base_directory`, `publish_directory`, `install_command`, `build_command`, `start_command` — they're typed but never passed in.

## Repro

Coolify v4 / coolify-mcp `2.x`. From a real session today:

```jsonc
// MCP create_key with everything I needed
{
  "action": "create_key",
  "name": "i-access-api",
  "project_uuid": "...",
  "server_uuid": "...",
  "destination_uuid": "...",
  "git_repository": "git@github.com:org/i-access.git",
  "git_branch": "main",
  "private_key_uuid": "...",
  "build_pack": "dockerfile",
  "ports_exposes": "3000",
  "domains": "https://i-access-api.example.com",
  "instant_deploy": false,
  "health_check_enabled": true,        // schema-valid, dropped on create
  "health_check_path": "/health",      // schema-valid, dropped on create
  "health_check_port": 3000,           // schema-valid, dropped on create
  "health_check_start_period": 60      // schema-valid, dropped on create
  // dockerfile_location: "/apps/api/Dockerfile" — would be schema-rejected
}
```

`get_application` afterwards: `dockerfile_location: null`, `health_check_enabled: false`, `health_check_path: "/"`. The same fields applied via raw `PATCH /api/v1/applications/{uuid}` work end-to-end (I confirmed against this Coolify instance).

## Suggested fix

Two small, independent changes; happy to open a PR if either looks good:

1. Extend the zod schema in the `application` tool with the missing build-config fields:
   `dockerfile_location`, `dockerfile_target_build`, `base_directory`, `publish_directory`, `install_command`, `build_command`, `start_command`, `watch_paths`. They already exist on `UpdateApplicationRequest`, so this fixes update by itself.

2. In the four `create_*` handlers, forward the health check fields and the build-config fields from §1 into the `createApplication*` calls. The `Create…Request` interfaces in `src/types/coolify.ts` need the same fields added (they're already on the equivalent update type).

I'd keep the existing hand-pick pattern rather than switching `create_*` to a spread, so the schema stays authoritative and fields surface in tool docs.

Storage / persistent volumes are out of scope here — covered by #172 which is already open.

Reproduced on a live Coolify v4 instance behind `coolify.example.com`; the underlying API accepts every one of these fields on PATCH today.

kashik0i:
Closes #178.

## What this changes

Two related gaps in the `application` tool, both surfacing when a caller passes valid Coolify fields and finds them silently dropped:

**1. Schema gap.** Eight build-config fields were absent from the zod schema for the `application` tool: `dockerfile_location`, `dockerfile_target_build`, `base_directory`, `publish_directory`, `install_command`, `build_command`, `start_command`, `watch_paths`. Zod's default behavior strips unknown keys without error, so an MCP caller could set `dockerfile_location` on `update`, get no feedback, and see no effect on the app. Added all eight to the schema.

**2. Handler gap.** The 12 `health_check_*` fields were already declared in the schema, but the four `create_*` hand-pick handlers never forwarded them. Same story for the build-config fields (which are typed on `CreateApplicationPublicRequest` but never passed through). Result: `create_key` with `health_check_enabled: true, health_check_path: '/health'` produced an app with healthcheck off and default `/` path. Added explicit forwards in `create_public`, `create_github`, `create_key`, and `create_dockerimage`, keeping the existing hand-pick pattern so the schema stays authoritative for tool docs.

## Two asymmetries confirmed against Coolify source (`coollabsio/coolify` `main`)

Posting these because they cut against the initial review summary in a meaningful way and shape this PR:

**`create_dockerimage` only forwards `health_check_*` (not build-config).** Coolify's `/applications/dockerimage` request-body schema in `openapi.yaml` does not accept `base_directory`/`publish_directory`/`install_command`/`build_command`/`start_command`/`watch_paths`/`dockerfile_location` — the endpoint is for pre-built registry images and has no build step. Wiring those into `create_dockerimage` would silently send dead weight. Handler `create_dockerimage` forwards healthcheck only.

**`dockerfile_target_build` is UPDATE-only.** You raised this in your comment — `dockerfile_target_build` isn't in `openapi.yaml`'s request bodies or any `Create*Request` interface. Verified against `app/Http/Controllers/Api/ApplicationsController.php` directly:

- `removeUnnecessaryFieldsFromRequest()` runs **before** validation, using two separate `$allowedFields` arrays.
- CREATE allowlist (line 1014): does **not** include `dockerfile_target_build` → silently stripped on every `create_*`.
- UPDATE allowlist (line 2497): **does** include `dockerfile_target_build` → accepted on PATCH.

The shared `sharedDataApplications()` validator declares a rule for the field, but on create paths the request never reaches it. So:

- The zod schema includes `dockerfile_target_build` so the `update` action can forward it via the existing spread.
- No `create_*` handler forwards it.
- Its presence in the schema for create paths is harmless (zod accepts; handler hand-pick simply doesn't reference it). Documented in `CLAUDE.md` so future maintainers don't add the forward.

Coolify's `openapi.yaml` request bodies are an incomplete projection of the real allowlists; full truth lives in the controller `$allowedFields` arrays. Captured this in `CLAUDE.md` as a gotcha so it doesn't bite the next contributor.

## Tests

Following the shape of #174 / #162:

- `src/__tests__/coolify-client.test.ts`: per-method forwarding tests for `createApplicationPublic`, `createApplicationPrivateGH`, `createApplicationPrivateKey`, `createApplicationDockerImage`, plus an `updateApplication` test for `dockerfile_target_build`.
- `src/__tests__/mcp-server.test.ts`: a new `describe('application tool handler')` block exercising each `create_*` action's forward via `_registeredTools['application'].handler(...)` + `jest.spyOn` (matches the `env_vars tool handler` pattern from #174). Includes a dedicated test asserting `create_dockerimage` drops `base_directory`/`install_command`/`dockerfile_location` (they should NOT reach the client).

`npm test`: 307 passed.
`npx tsc --noEmit`: clean.
`npm run lint`: clean (no new warnings).
`npx prettier --check`: clean.

## Live smoke test

Built the dist and ran against a live Coolify v4 instance:

- `create_public` with `build_pack: dockerfile`, `dockerfile_location`, all 7 other build-config fields, and 4 `health_check_*` fields → GET round-trips every field. (`health_check_port` is returned as `string`, which is Coolify's column-level type coercion at the DB layer, not a client behaviour.)
- `update` with `dockerfile_target_build: 'production'` + an updated `dockerfile_location` → GET confirms both persisted.
- DELETE cleanup succeeded.

## Out of scope (happy to send as follow-up PRs)

A few other application-tool gaps surfaced during the same investigation. Same shape, smaller diffs each:

- Static-site flags (`is_static`, `is_spa`, `static_image`)
- HTTPS/domain flags (`is_force_https_enabled`, `autogenerate_domain`, `force_domain_override`, `redirect`)
- HTTP basic auth (`is_http_basic_auth_enabled`, `http_basic_auth_username`, `http_basic_auth_password`)
- Resource limits (`limits_memory`, `limits_memory_swap`, `limits_cpus`, etc.)
- Pre/post-deploy hooks (`post_deployment_command(_container)`, `pre_deployment_command(_container)`)
- Per-provider manual webhook secrets (`manual_webhook_secret_github/gitlab/bitbucket/gitea`)
- Build-server toggles (`use_build_server`, `connect_to_docker_network`)
- Container label preservation flags (`is_container_label_escape_enabled`, `is_preserve_repository_enabled`)
- Docker Compose extras (`docker_compose_location`, `docker_compose_custom_*_command`, `docker_compose_domains`)
- Read-side normalization of `health_check_port` (string from Coolify → number on MCP layer)


StuMason (approved):
Thanks @kashik0i — this is the gold standard for this kind of fix:

- Verified against Coolify's actual controller source (`ApplicationsController.php` lines 1014 vs 2497), not just OpenAPI — and correctly caught that `dockerfile_target_build` is UPDATE-only, which is the opposite of what I'd assumed when triaging #178. The asymmetry now lives in CLAUDE.md so the next contributor doesn't have to rediscover it.
- `create_dockerimage` correctly drops build-config (pre-built images have no build step) — the test that asserts `not.toHaveProperty('base_directory')` etc. locks that intentional behaviour in place.
- Handler forwarding is real, not just zod schema additions — that was the pre-existing gap from #162 and this closes it cleanly.
- Test pattern matches the `_registeredTools[...].handler(...)` style from #174 — behavioural not nominal.
- Live smoke-tested against a real Coolify v4 instance with round-trip GET verification.

Out-of-scope follow-up list at the bottom of the PR description is useful — happy to take any of those as separate PRs whenever you've got the bandwidth.

Approving. Closes #178.

Issue #159, opened 2026-04-19
env_vars tool result leaks plaintext value and real_value


daniel-rudaev:
### Issue

The `env_vars` / `bulk_env_update` tool (and the underlying GET /applications/{uuid}/envs endpoint) returns the full env var record in the MCP `tool_result`, including `value` and `real_value` in plaintext. When an AI agent calls the tool to manage environment variables, the value enters the agent's conversation transcript and session storage — defeating the operator's confidentiality expectation for secret-class env vars.

### Reproduction

1. Have a Coolify app with env var `FOO=secret123`
2. Call the MCP tool that lists / creates / updates envs on that app
3. Observe the tool_result JSON contains `"value": "secret123"` and `"real_value": "secret123"`
4. The AI agent now has the plaintext value in context + transcript

### Impact

Any agent using coolify-mcp with write permissions effectively also reads back every value it touches (plus every value already on the app, via list). For secret-bearing env vars (API keys, database URLs, tokens), this is a leak into:

- The agent's context window (in-memory)
- The agent's session transcript storage (on disk, per the agent's host)
- Any logs the agent itself produces

### Proposed fix

Three options, preference order:

1. **Default: mask values in MCP responses.** Return `"value": "***"` (or omit entirely) for any env var created or read via the MCP. Preserve the metadata (key, uuid, flags, timestamps) so operations on the record still work.
2. **Opt-in reveal:** add a `--reveal` / `reveal=true` parameter the caller must set explicitly. Default masked.
3. **Per-field sensitivity flag:** if Coolify's env var record has or could have an `is_secret` flag, mask only those. Less safe (defaults allow leaks), but lower-impact on existing callers.

Option 1 matches what `n8n`'s public API does for password-type credential fields and what `1Password`'s CLI does by default.

### Context

We're building a secret-capture tool that routes user-typed secrets to destinations including Coolify. To avoid the value entering our agent's transcript, we bypassed the MCP and called Coolify's REST directly with `curl ... -o /dev/null` on writes. Would prefer to use the MCP if responses were masked.

Our adapter: https://github.com/D1DX/secret-capture-skill/blob/main/scripts/adapters/coolify.sh

Happy to submit a PR — just want to align on the approach first.

StuMason:
[...]

PR #206, opened 2026-05-24
fix(system): list_resources essential projection, secret masking, is_build_time typo cleanup

### LINKED ISSUE #203: system list_resources returns full Coolify payload by default — same bug class as pre-#158 listApplicationDeployments
## Summary

`system({ action: 'list_resources' })` (introduced in v2.11.0 / #172) typed as `Promise<ResourceListItem[]>` where `ResourceListItem` is `{ uuid, name, type, status? }`, but the underlying Coolify `/api/v1/resources` endpoint actually returns the full nested resource record (~95 fields per item including build/healthcheck/limits/git/docker-compose config). The TypeScript type is runtime-erased, so the client just casts the bloated response without projecting it. This is the same bug class that PR #158 fixed for `listApplicationDeployments` (was typed `Deployment[]`, returned `{ count, deployments }` envelope with embedded logs blobs).

## Reproduction

Against a Coolify v4 instance with a modest 33 apps/services:

| Metric | Value |
|---|---|
| Tool call | `system({ action: 'list_resources' })` |
| HTTP endpoint | `GET /api/v1/resources` |
| Item count | 33 |
| Response size | **537,930 bytes (~525 KB)** |
| Avg bytes / item | ~16,300 |
| Keys per item | ~95 (vs. 4 promised by `ResourceListItem`) |

Sample of fields actually returned that the type does not declare (truncated):

```
additional_networks_count, additional_servers, additional_servers_count,
base_directory, build_command, build_pack, compose_parsing_version,
config_hash, created_at, custom_docker_run_options, custom_healthcheck_found,
custom_labels, custom_network_aliases, custom_nginx_configuration, deleted_at,
description, destination, destination_id, destination_type, docker_compose,
docker_compose_custom_build_command, docker_compose_custom_start_command,
docker_compose_domains, docker_compose_location, docker_compose_raw,
docker_registry_image_name, docker_registry_image_tag, dockerfile,
dockerfile_location, dockerfile_target_build, environment_id, fqdn,
git_branch, git_commit_sha, git_full_url, git_repository,
health_check_command, health_check_enabled, health_check_host,
health_check_interval, health_check_method, health_check_path,
health_check_port, health_check_response_text, health_check_retries,
health_check_return_code, health_check_scheme, health_check_start_period,
health_check_timeout, health_check_type, ...
```

## Impact

- **MCP token budget**: 525 KB blows past typical MCP client/LLM context limits on a single tool call. The result has to be spilled to disk by the host, after which the LLM cannot meaningfully consume it. Effectively renders `list_resources` unusable for any agentic workflow on instances with non-trivial resource counts.
- **Type lie**: any TypeScript consumer destructuring `ResourceListItem` fields works (lucky), but anyone relying on the typed surface to reason about response size or field set is wrong-at-runtime.
- **Scales linearly with instance size**: 33 resources is small. A larger instance (100+ apps/services) easily exceeds 1.5 MB on this single call.

## Proposed fix (mirrors PR #158)

1. Introduce a true runtime essential projection:

   ```ts
   export interface ResourceListItemEssential {
     uuid: string;
     name: string;
     type: 'server' | 'application' | 'database' | 'service' | string;
     status?: string;
   }
   ```

   (i.e. what `ResourceListItem` already claims to be, but now actually enforced at the API boundary in `listResources()` by mapping over the response.)

2. Update `listResources()` to project by default and accept an opt-in:

   ```ts
   async listResources(
     options?: { include_full?: boolean }
   ): Promise<ResourceListItemEssential[] | FullResourceListItem[]>
   ```

3. Expose `include_full?: boolean` on the `system` tool's `list_resources` action. Default `false` (essential projection). Set `true` to opt back into the raw Coolify response.

4. Add an integration test that asserts the default response shape contains only the essential keys, guarding against future regressions.

## Notes

- Naming `include_full` (rather than `include_logs`) because the heavy data on `/resources` is not a single field but the full record. Pattern parity with PR #158's `include_logs: true` opt-in.
- A separate issue will cover secret-masking on the `include_full` path — the raw Coolify response contains `manual_webhook_secret_*` and `http_basic_auth_password` fields that should be masked by default (analog of the v2.9.0 `env_vars` masking from #159/#182).
- Happy to send the PR.

### LINKED ISSUE #204: system list_resources exposes webhook secrets and basic-auth credentials in MCP responses — apply v2.9.0 env_vars masking pattern
## Summary

`system({ action: 'list_resources' })` (v2.11.0 / #172) returns the full Coolify `/api/v1/resources` response, which embeds several per-resource secrets in plaintext on every application record:

- `manual_webhook_secret_github`
- `manual_webhook_secret_gitlab`
- `manual_webhook_secret_gitea`
- `manual_webhook_secret_bitbucket`
- `http_basic_auth_password`

These are the HMAC secrets used to validate inbound deploy webhooks (forging one lets an attacker trigger a deploy on that repo with controlled payload), and the HTTP basic-auth password used to gate front-of-app access. They are currently returned to any MCP client / LLM that calls `list_resources`, with no opt-in.

This is the same exposure class that PR #159 / #182 fixed for `env_vars` in v2.9.0 — `value` and `real_value` were being shipped in plaintext and are now masked by default with an explicit `reveal: true` opt-in. The same posture should apply here.

## Reproduction

On any Coolify v4 instance with at least one git-deployed application:

```bash
# via MCP
system({ action: 'list_resources' })

# equivalent HTTP
curl -H "Authorization: Bearer $COOLIFY_TOKEN" \
  https://<coolify>/api/v1/resources | jq '.[] |
  { uuid, manual_webhook_secret_github, manual_webhook_secret_gitlab,
    manual_webhook_secret_gitea, manual_webhook_secret_bitbucket,
    http_basic_auth_password }'
```

Every application row has the four `manual_webhook_secret_*` strings populated (they are auto-generated by Coolify even when the user only uses one provider) plus the basic-auth password if `is_http_basic_auth_enabled`.

## Impact

- **Webhook-secret leakage**: an MCP client / LLM that has been granted permission to call `list_resources` (a "read-only enumeration" tool by name and description) silently also learns four HMAC secrets per app. A malicious or compromised LLM agent can use them to forge deploy webhooks. The user granting "list resources" permission almost certainly did not intend to also grant "exfiltrate every webhook signing key".
- **Token-equivalent escalation**: `manual_webhook_secret_github` (and friends) are independent of the Coolify API token — an attacker who pivots through them retains the ability to trigger deploys even after the Coolify API token is rotated.
- **Basic-auth password**: same pattern. Read-only enumeration shouldn't divulge the password.
- The Coolify API token holder is naturally trusted with this data, but the point of the MCP layer is to expose a *narrower* surface to LLM clients than the raw token — same reasoning as #159 / #182.

## Proposed fix

Apply the v2.9.0 pattern exactly:

1. Define a hard-coded list of sensitive fields per resource type:

   ```ts
   const SENSITIVE_RESOURCE_FIELDS = [
     'manual_webhook_secret_github',
     'manual_webhook_secret_gitlab',
     'manual_webhook_secret_gitea',
     'manual_webhook_secret_bitbucket',
     'http_basic_auth_password',
   ] as const;
   ```

2. In `listResources()`, walk each item and replace each sensitive field with `'***'` unless the caller passed `reveal: true`. Apply at the API boundary in the client (not the tool handler) so any other code path that calls `listResources()` also gets masking by default.

3. On the `system` tool, accept `reveal?: boolean` on the `list_resources` action and thread it through. Match the existing `env_vars({ reveal })` ergonomics exactly so callers don't have to learn a new flag name.

4. Add a test that asserts default response has `***` in each sensitive field and that `reveal: true` round-trips the original value.

## Notes

- This is a defense-in-depth fix at the MCP boundary; the underlying Coolify API obviously still returns these fields. The MCP layer is the right place to apply masking because that is where the trust boundary narrows.
- Pairs with the essential-projection work in #203 — once `list_resources` defaults to the essential projection, the leaked secret fields aren't in the default response at all. Masking still matters because it protects callers who legitimately pass `include_full: true`.
- Happy to send the PR.

### LINKED ISSUE #205: is_build_time typo bleed-in to v2.11.0 — test mock + CHANGELOG only, runtime code is correct
## Summary

PR #174 (v2.9.0) renamed the env-var build-time flag from `is_build_time` to `is_buildtime` everywhere — Coolify's API rejects the underscored form with HTTP 422 on single endpoints and silently ignores it on bulk endpoints.

PR #172 (v2.11.0, `feat: 7 new tools`) was authored against pre-#174 main and, during the merge into post-#174 main, two `is_build_time` references slipped back into the codebase in non-runtime locations:

1. **Test mock** — `src/__tests__/coolify-client.test.ts:4335`, inside the new `listDatabaseEnvVars` test added for the `database` env_vars resource:

   ```ts
   const mockEnvs = [
     {
       id: 1,
       uuid: 'env-1',
       key: 'DB_HOST',
       value: 'localhost',
       is_build_time: false,   // ← should be `is_buildtime`
       ...
     },
   ];
   ```

   The mock just round-trips through `expect(result).toEqual(mockEnvs)`, so the test passes despite the typo, but the mock no longer matches the `EnvironmentVariable` type definition in `src/types/coolify.ts:442-489`.

2. **CHANGELOG** — `[2.11.0]` section, `env_vars expanded` bullet:

   > adds `database` resource, `bulk_update` action, plus `is_build_time`, `is_preview`, `data[]` params on the existing actions.

   The underscored name is incorrect; the actual field added on the schema is `is_preview` (and `data[]`), not `is_build_time`. `is_buildtime` has been the correct name since v2.9.0.

## Verification that runtime code is unaffected

Audit on v2.11.0 source:

| Surface | `is_buildtime` | `is_build_time` |
|---|---|---|
| `src/types/coolify.ts` | 4 occurrences | 0 |
| `src/lib/coolify-client.ts` | 3 occurrences | 0 |
| `src/lib/mcp-server.ts` | 13 occurrences | 0 |
| `src/__tests__/coolify-client.test.ts` | (many) | **1** (this issue) |
| `dist/` (excluding tests) | 65 occurrences | 0 |

So no functional regression — just doc and test-mock cleanup.

## Proposed fix

- `src/__tests__/coolify-client.test.ts:4335` — change `is_build_time` → `is_buildtime`.
- `CHANGELOG.md` `[2.11.0]` line — change `is_build_time` → `is_buildtime` (or drop the field name entirely; the new params on `env_vars` expansion are really `is_preview` and `data[]`, since `is_buildtime` has been on the schema since v2.9.0).

Trivial. Happy to bundle into the PR for #203 / #204 since it's a one-line fix in two files and touches the same release.

kashik0i:
Closes #203, #204, #205.

This PR addresses three issues that surfaced post-v2.11.0 when wiring up the new `system` tool against a real Coolify instance with 30+ resources. All three are scoped to `list_resources` plus a small docs/test-mock typo bleed-in from #172. Mirrors the same opt-in / masking patterns shipped previously by #158 (`include_logs`) and #159 / #182 (`env_vars` masking with `reveal`).

### 1. `fix(system): default list_resources to essential projection, add include_full opt-in` — closes #203

Previously typed as `Promise<ResourceListItem[]>` but actually returned the full Coolify `/api/v1/resources` payload (~95 fields per row, ~16 KB per item). On a moderate instance with 33+ resources this exceeded 500 KB on a single call, blowing MCP token budgets and effectively making `list_resources` unusable for LLM-driven workflows.

Now defaults to a true `{ uuid, name, type, status? }` projection applied at the API boundary in `listResources()`. Set `include_full: true` to opt back into the raw Coolify payload.

**Live verification on a Coolify v4 instance (34 resources):**

| Call | Response size | Item keys |
|---|---|---|
| `list_resources` (default) | **3,749 bytes** | 4 (`uuid`/`name`/`type`/`status`) |
| `list_resources include_full=true` | ~105 KB | ~95 (raw Coolify) |

~143× smaller on the default path.

### 2. `fix(security): mask webhook secrets and basic-auth password in list_resources` — closes #204

The full `/api/v1/resources` response embeds five sensitive fields on every application row:

- `manual_webhook_secret_github`
- `manual_webhook_secret_gitlab`
- `manual_webhook_secret_gitea`
- `manual_webhook_secret_bitbucket`
- `http_basic_auth_password`

The first four are HMAC signing keys for inbound deploy webhooks — forging one lets an attacker trigger a deploy on that repo independent of the Coolify API token. The basic-auth password gates front-of-app access.

Before this change, any MCP client / LLM granted permission to call `system list_resources` with `include_full: true` silently received every one of these. The user almost certainly did not intend to grant secret-exfiltration when granting a "read-only enumeration" tool.

Applies the v2.9.0 `env_vars` masking pattern (#159 / #182) exactly: replace each sensitive field with `'***'` at the API boundary in `listResources()`. Pass `reveal: true` alongside `include_full: true` to round-trip plaintext. Masking is applied at the client method so any other code path that calls `listResources()` also inherits the protection.

**Live verification on the same instance (11 git-deployed apps):**

| Call | Secrets `***` | Secrets plaintext |
|---|---|---|
| `include_full=true` (default `reveal=false`) | **11/11** | 0 |
| `include_full=true reveal=true` | 0 | **11/11** |

### 3. `docs: fix is_build_time typo bleed-in from #172` — closes #205

PR #174 renamed the env-var build-time flag from `is_build_time` to `is_buildtime` everywhere (Coolify's API rejects the underscored form with HTTP 422 on single endpoints and silently ignores it on bulk endpoints). PR #172 was authored against pre-#174 main and during the merge into post-#174 main, two `is_build_time` references slipped back in:

1. Test mock at `src/__tests__/coolify-client.test.ts:4335` (mockFetch is typed as `any` so TypeScript didn't catch the drift from the `EnvironmentVariable` type definition)
2. `CHANGELOG [2.11.0]` `env_vars expanded` bullet

Runtime code was already correct — this is doc + test-mock cleanup only.

## Breaking change (typed-only)

- **`listResources` signature** (#203, #204): `Promise<ResourceListItem[]>` → `Promise<ResourceListItem[] | ResourceListItemFull[]>` with new optional `options?: { include_full?: boolean; reveal?: boolean }` parameter. Programmatic consumers of `@masonator/coolify-mcp` will need to widen their result type (or narrow at the call site with `include_full`). Note: the previous type was wrong-at-runtime against real Coolify (claimed 4 fields, returned ~95), so any caller that worked end-to-end was already accommodating the bloated shape.

## Verification

- `pnpm build` — passes (tsc clean).
- `pnpm lint` — 0 errors, 4 pre-existing warnings (unrelated `any` types in error handlers).
- `pnpm test` — **356 tests pass** (was 348 before; +8 new tests covering essential projection, masking, reveal, null-secret edge case, default-projection-with-reveal no-op).
- Coverage: 98.09% statements / 99.53% functions.
- Live smoke against a real Coolify v4 instance (34 resources / 11 apps) confirmed all three behaviors as documented above.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,18 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Fixed
+
+- **`system list_resources` returns essential projection by default** (#203) — previously typed as `Promise<ResourceListItem[]>` but actually returned the full Coolify `/api/v1/resources` payload (~95 fields per row, ~16 KB per item, 500+ KB on instances with 30+ resources) which blew MCP token budgets and made the tool unusable for LLM-driven workflows. Now defaults to a true `{ uuid, name, type, status? }` projection applied at the API boundary. Set `include_full: true` to opt back into the raw Coolify payload. Mirrors the `include_logs` opt-in pattern from #158.
+
+### Changed (breaking, typed-only)
+
+- **`listResources` signature** (#203): `Promise<ResourceListItem[]>` → `Promise<ResourceListItem[] | ResourceListItemFull[]>` with new optional `options?: { include_full?: boolean }` parameter. Programmatic consumers of `@masonator/coolify-mcp` will need to widen their result type (or narrow at the call site with `include_full`). Note: the previous type was wrong-at-runtime against real Coolify (claimed 4 fields, returned ~95), so any caller that worked end-to-end was already accommodating the bloated shape.
+
+### Added (types)
+
+- **`ResourceListItemFull` type** — `ResourceListItem & Record<string, unknown>`. Surfaces only when `include_full: true` is passed; documents that the full Coolify row carries arbitrary additional fields beyond the typed essentials.
+
 ## [2.11.0] - 2026-05-18
 
 ### Added (#172, thanks @opastorello)
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 ### Fixed
 
 - **`system list_resources` returns essential projection by default** (#203) — previously typed as `Promise<ResourceListItem[]>` but actually returned the full Coolify `/api/v1/resources` payload (~95 fields per row, ~16 KB per item, 500+ KB on instances with 30+ resources) which blew MCP token budgets and made the tool unusable for LLM-driven workflows. Now defaults to a true `{ uuid, name, type, status? }` projection applied at the API boundary. Set `include_full: true` to opt back into the raw Coolify payload. Mirrors the `include_logs` opt-in pattern from #158.
+- **`is_build_time` typo bleed-in from #172** (#205) — test mock at `src/__tests__/coolify-client.test.ts:4335` and the `[2.11.0]` CHANGELOG entry both referenced the wrong underscored field name (`is_build_time`). The correct name is `is_buildtime` (one word) per #174 / v2.9.0; runtime code was already correct, this is doc + test-mock cleanup only.
 
 ### Changed (breaking, typed-only)
 
@@ -33,7 +34,7 @@ Net tool count: 38 → 42 (after consolidation; original PR proposed 45 before r
 - **`scheduled_tasks` tool** — list/create/update/delete/list_executions for application or service. Consolidated.
 - **`hetzner` tool** — list_locations, list_server_types, list_images, list_ssh_keys, create_server against the Coolify Hetzner cloud-provider endpoints. Requires a configured cloud-provider token UUID. The Coolify Hetzner routes are auth-scope-gated; the wiring is correct but the calling user needs the right token.
 - **`system` tool** — health, list_resources, enable_api, disable_api against the instance. Consolidates what would have been three separate tools.
-- **`env_vars` expanded** — adds `database` resource, `bulk_update` action, plus `is_build_time`, `is_preview`, `data[]` params on the existing actions.
+- **`env_vars` expanded** — adds `database` resource, `bulk_update` action, plus `is_preview` and `data[]` params on the existing actions. (`is_buildtime` and `is_runtime` have been on the schema since v2.9.0 / #174 — they are not new in this release.)
 - **`github_apps` expanded** — adds `list_repos`, `list_branches` actions.
 - **`database_backups` expanded** — adds `delete_execution` action.
 - **`ResourceListItem` type** — concrete interface for `/api/v1/resources` responses with `uuid`, `name`, `type`, optional `status`. Replaces `Promise<unknown>` on the client method (no-implicit-any policy compliance).
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,6 +15,7 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **`system list_resources` returns essential projection by default** (#203) — previously typed as `Promise<ResourceListItem[]>` but actually returned the full Coolify `/api/v1/resources` payload (~95 fields per row, ~16 KB per item, 500+ KB on instances with 30+ resources) which blew MCP token budgets and made the tool unusable for LLM-driven workflows. Now defaults to a true `{ uuid, name, type, status? }` projection applied at the API boundary. Set `include_full: true` to opt back into the raw Coolify payload. Mirrors the `include_logs` opt-in pattern from #158.
 - **`is_build_time` typo bleed-in from #172** (#205) — test mock at `src/__tests__/coolify-client.test.ts:4335` and the `[2.11.0]` CHANGELOG entry both referenced the wrong underscored field name (`is_build_time`). The correct name is `is_buildtime` (one word) per #174 / v2.9.0; runtime code was already correct, this is doc + test-mock cleanup only.
+- **`toResourceListItemEssential` guards `status` with `typeof`** — replaces an unchecked `item.status as string` cast. If a future Coolify version emits `status` as an object/null, the field is now dropped from the essential projection instead of silently violating the `ResourceListItem` interface.
 
 ### Changed (breaking, typed-only)
 
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,13 +7,17 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **`list_resources` masks webhook secrets and basic-auth credentials by default** (#204) — when `system({ action: 'list_resources', include_full: true })` is called, the per-resource fields `manual_webhook_secret_github` / `manual_webhook_secret_gitlab` / `manual_webhook_secret_gitea` / `manual_webhook_secret_bitbucket` and `http_basic_auth_password` are now replaced with `'***'` so an MCP client / LLM granted "list resources" cannot silently exfiltrate webhook HMAC signing keys or front-of-app passwords. Pass `reveal: true` alongside `include_full: true` to round-trip plaintext. Mirrors the v2.9.0 `env_vars` masking posture from #159 / #182. Applied at the API boundary in `listResources()` so any other code path also inherits masking by default.
+
 ### Fixed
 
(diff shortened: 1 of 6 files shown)


StuMason:
@claude please review this PR end-to-end against the repo's CLAUDE.md conventions. Focus on:

- Whether the masking pattern matches the v2.9.0 `env_vars` precedent (#159 / #182) exactly — same sentinel, same `reveal` semantics, applied at the same layer.
- Whether the breaking `listResources` signature change is acceptable for a patch (the previous return type was wrong-at-runtime).
- Test coverage gaps for edge cases (e.g. resources missing `name`/`type`, `include_full: true` on an empty list).
- Any other quality issues.

Just pushed a prettier auto-fix commit (`ea7695b`) on the contributor's behalf — pure formatting, no semantic changes.

StuMason (approved):
Thanks @kashik0i!

StuMason (approved):
Approving — already reviewed inline, all 5 checks green, follow-up #209 opened for the nested-payload audit. Thanks @kashik0i for the well-scoped, well-documented fix and the live-instance verification.

PR #212, opened 2026-05-31
Add pull_latest parameter to restart action for services

NeilHanlon:
The Coolify API supports `?latest=true` on the service restart endpoint, which pulls fresh Docker images before restarting. This is the API equivalent of the "Pull Latest Images & Restart" button in the Coolify UI.

Add an optional `pull_latest` boolean parameter to the `control` tool. When `resource=service`, `action=restart`, and `pull_latest=true`, appends `?latest=true` to the request URL. No behavior change for other resource types or when omitted.

## Summary

- Add optional `pull_latest` boolean to `control` tool schema (services only)
- `restartService()` appends `?latest=true` query string when set
- No behavior change for applications, databases, or when `pull_latest` is omitted/false

## Changes

- `src/lib/coolify-client.ts` — `restartService(uuid, pullLatest)` accepts optional second param
- `src/lib/mcp-server.ts` — `control` tool schema gains `pull_latest: z.boolean().optional()`, passed through for service restarts
- `src/__tests__/coolify-client.test.ts` — Two new tests: `pull_latest=true` appends `?latest=true`, `pull_latest=false` omits it

## Test plan

- [x] Tests added/updated
- [x] Manually tested locally

## Checklist

- [ ] Documentation updated if needed

### DIFF
--- a/src/__tests__/coolify-client.test.ts
+++ b/src/__tests__/coolify-client.test.ts
@@ -2425,6 +2425,30 @@ describe('CoolifyClient', () => {
         expect.objectContaining({ method: 'GET' }),
       );
     });
+
+    it('should restart a service with pull_latest', async () => {
+      mockFetch.mockResolvedValueOnce(mockResponse({ message: 'Restarted' }));
+
+      const result = await client.restartService('test-uuid', true);
+
+      expect(result).toEqual({ message: 'Restarted' });
+      expect(mockFetch).toHaveBeenCalledWith(
+        'http://localhost:3000/api/v1/services/test-uuid/restart?latest=true',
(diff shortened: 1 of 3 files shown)


StuMason:
Thanks for this, @NeilHanlon — clean, focused, and exactly the right shape. 🙌

The backwards-compatible default (`pullLatest = false`), the wiring through the `control` tool with a clear description, and the two tests covering both the `?latest=true` and default paths made this an easy yes.

The only thing standing in the way was a prettier formatting nit (a couple of lines over 100 chars) plus needing a sync with `main` after a dependency-overrides change landed. I've pushed both to your branch — CI is green across Node 20/22/24 and all 363 tests pass. Merging now.

PR #220, opened 2026-06-11
PoC: frozenminisearch for docs search (lower MCP heap)

yoch:
## Summary

This PR is an optional PoC: replace `minisearch` with [`@yoch/frozenminisearch`](https://github.com/yoch/frozenminisearch) for the documentation search path.

`frozenminisearch` is a read-only, Node-only drop-in with the same `search()` / BM25 / prefix / fuzzy API as MiniSearch 7, but with a much smaller memory footprint and optional binary snapshots (MSv5).

## Why this might fit this project

- Docs index is built once at startup (`addAll`), then only queried — no `remove`/`discard`.
- MCP processes are often long-lived; lower retained heap helps users running multiple servers.

## Benchmark (scaled corpus, Node v24.16)

Corpus: unit-test fixture chunks ×500 (4,500 documents) — same field layout and search options as production.

| | MiniSearch | FrozenMiniSearch |
|---|------------|------------------|
| Index heap (retained, post-GC) | 8.06 MB | 0.52 MB (~93.5% less) |
| Search parity (ranking) | — | identical top results on sample queries |

Score values can differ slightly (Float32 BM25) but result ordering matches on all tested queries.

## Migration surface

- Swap index construction: `new MiniSearch(opts).addAll(docs)` → `FrozenMiniSearch.fromDocuments(docs, opts)`
- `search()` call sites unchanged
- Requires Node >= 22.15 (zstd in `node:zlib`; bumped in `engines`)

## Notes

- No pressure to merge — happy to close if out of scope.
- Compatibility notes tracked in [frozenminisearch friction log](https://github.com/yoch/frozenminisearch/blob/main/dev/drop-in-friction-log.md).

## Test plan

- [x] Existing unit tests pass (`npm test`, 361 tests)
- [ ] Manual `search_docs` returns equivalent top results for sample queries in your environment

Made with [Cursor](https://cursor.com)

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -10,7 +10,7 @@
       "license": "MIT",
       "dependencies": {
         "@modelcontextprotocol/sdk": "^1.23.0",
-        "minisearch": "^7.2.0",
+        "@yoch/frozenminisearch": "file:../../..",
         "zod": "^4.3.5"
       },
       "bin": {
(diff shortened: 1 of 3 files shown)


StuMason:
Thanks for putting this together, @yoch — and for being upfront that it's an optional PoC. The code-level change is genuinely clean: the `FrozenMiniSearch.fromDocuments(...)` swap maps neatly onto the existing MiniSearch setup, and `frozenminisearch` being a drop-in BM25 API makes the diff pleasantly small.

After a proper look, though, I'm going to close this one rather than merge — a few blockers that together outweigh the benefit for this project:

1. **Node 20 drop.** `engines` moves `>=20` → `>=22.15`. `coolify-mcp` is a published package with users on Node 20 LTS (and the CI matrix tests 20/22/24), so dropping that line is a breaking change rather than a free win.
2. **Install path.** The committed lockfile resolves the dep to `file:../../..` (a local checkout) against a `^1.1.0` spec in `package.json`, so `npm ci` fails validation — that's the red CI here.
3. **Marginal gain for this workload.** The docs index is built at runtime from `llms-full.txt` (a few hundred KB) inside a long-running process, and the PoC uses `fromDocuments(...)` rather than a prebuilt frozen snapshot — so it isn't leveraging frozenminisearch's main advantage, and the heap delta for an index this size is small. There's also no open issue reporting memory pressure that this would address.

None of that is a knock on the library — it just isn't a compelling trade for *this* search path right now. If you'd like to revisit it, the bar would be: keep Node 20 support, install from the registry, and include before/after memory numbers for an index of this size (ideally using the binary-snapshot path, which is where the real win would be). Appreciate the contribution either way. 🙏

Issue #232, opened 2026-07-02
deployment get with `lines` bypasses the essential projection — leaks server secrets and blows token budgets


StuMason:
## What happens

`deployment {action: get, uuid, lines: N}` switches from the slim `toDeploymentEssential()` projection to the **raw upstream payload** (`getDeployment(uuid, { includeLogs: true })`), and that raw Coolify response embeds the entire application graph: rendered docker_compose, custom_labels, webhook secrets, the destination **server** object with proxy config, and `server.settings` including **`logdrain_custom_config` (which contains a live bearer token for the log drain) and `sentinel_token`**.

Observed in production use (2026-07-02): `deployment {action: get, uuid: ..., lines: 5}` returned a ~78KB payload — 5 log lines wrapped in the full application + server + proxy + log-drain config. It blew the calling agent's tool-result cap, and the secrets above were in the body.

## Why it matters

1. **Security**: v2.12.0 deliberately masked webhook secrets in `list_resources` — this path undoes that class of protection. Any agent asking for deployment logs gets log-drain credentials and the sentinel token in context (and potentially into transcripts/telemetry).
2. **Token budget**: the whole point of `toDeploymentEssential` (coolify-client.ts ~line 300) is defeated the moment you ask for logs, which is the most common reason to call `get` at all.

## Where

- Handler: `src/lib/mcp-server.ts:1155` — the `lines` branch calls `getDeployment(uuid, { includeLogs: true })` and returns the raw object.
- Projection helper: `src/lib/coolify-client.ts` ~300 (`toDeploymentEssential`).

## Suggested fix

Always project through `toDeploymentEssential`, and when `lines` is set attach `logs` (via the existing `truncateLogs()` in `mcp-server.ts:70`) onto the *projection* — never return the raw upstream object. If any raw field is genuinely needed later, add it to the projection explicitly. Belt-and-braces: run the projection output through the same secret-masking used by `list_resources` so future upstream fields can't reintroduce a leak.

## Acceptance

- `deployment get` with `lines` returns: essential fields + `logs` (truncated) only; no `application.*`, no `server.*`, no compose, no tokens.
- A test asserting the response for a logs-included get contains no `logdrain`, `sentinel_token`, `manual_webhook_secret*`, or `docker_compose` keys.

PR #246, opened 2026-07-02
feat: add wait option to deploy tool (#238)

### LINKED ISSUE #238: deploy: add wait option to poll deployment to terminal status
Split from #237 (item 1).

## Motivation
A webhook deploy failed silently (build container OOM-killed, exit 255) while the old container kept serving — the app looked fine, the deploy was dead. The `deploy` tool returns a `deployment_uuid` and walks away, so an agent that deploys without watching ships "site 200 = deployed" false positives.

## Proposal
Add `wait: boolean` (default false) and optional `timeout_seconds` (default ~300) to the `deploy` tool:
1. Trigger deploy as today, capture `deployment_uuid`.
2. Poll `getDeployment(uuid)` every ~5s until terminal status (`finished` / `failed` / `cancelled`) or timeout.
3. On failure, include the tail of the build logs (reuse `truncateLogs`) so the caller sees *why* without a second call.
4. On timeout, return current status + the deployment UUID and a next-action hint to keep polling via `deployment get`.

Response must stay small: status, timings, and a bounded log tail on failure only — never the raw upstream payload (see #232).

## Acceptance
- `deploy {tag_or_uuid, wait: true}` returns final status; on failure includes a bounded log tail.
- Timeout path is explicit and actionable.
- Tests for success, failure, and timeout paths.

StuMason:
## Summary

Closes #238

`deploy` used to fire-and-forget: a build could fail (e.g. build container OOM-killed) while the old container kept serving, and the caller never found out — a deployed-looking site that's actually still on stale code.

Adds an opt-in `wait: boolean` (+ `timeout_seconds`, default 300) to the `deploy` tool:

- Triggers the deploy as before, then polls `getDeployment` every ~5s until a terminal status (`finished` / `failed` / `cancelled`) or the timeout elapses.
- **Success**: returns `status`, `deployment_uuid`, `commit`, timestamps, and `duration_seconds`.
- **Failure**: same, plus a bounded log tail (`truncateLogs`, last 30 entries / 10KB) — built from essential fields only, never the raw upstream `Deployment` object (which can carry `server`/`application` secrets — see #232).
- **Timeout**: current status + `timed_out: true` + a `next_action` hint to keep polling via `deployment get`.
- **No-wait path is byte-for-byte unchanged.**

### Multi-deployment tags

A tag can match several applications, so Coolify's `/deploy` response can contain multiple `deployment_uuid`s. This PR **watches only the first** and lists the rest under `additional_deployment_uuids` in the response, so the caller can poll them separately via `deployment get`. Watching all of them concurrently was considered but adds meaningful complexity (parallel polling, combined timeout/status semantics) for a case that's uncommon in practice; the explicit `additional_deployment_uuids` field keeps the behavior discoverable instead of silent.

### Side fix

`deployByTagOrUuid` was typed as `Promise<MessageResponse>` (`{ message: string }`), but Coolify's real `/deploy` response is `{ deployments: [{ message, resource_uuid, deployment_uuid }] }` (confirmed against `docs/openapi-chunks/deployments-api.yaml`). This was a latent "wrong-at-runtime" type (same class of bug as #158), and had to be fixed to extract the deployment uuid(s) to poll. New `DeployTriggerResponse` type added; old `{ message }`-only shape still works for backwards compatibility.

### Test mockability

Poll sleeps use an isolated `sleep()` helper so tests drive them with `jest.useFakeTimers()` + `jest.advanceTimersByTimeAsync()` rather than real waits.

## Test plan

- [x] `npm test` — 369 passed (5 new: success/failure/timeout/no-wait/multi-deployment-first-only)
- [x] `npm run lint` — clean (pre-existing warnings only, no errors)
- [x] `npx prettier --check .` — clean
- [x] `npm run build` — passes
- [x] Verified failure-path test asserts the response never contains `server`/`application`/`private_key`/`env_secret` fields from a raw upstream payload

### DIFF
--- a/src/__tests__/coolify-client.test.ts
+++ b/src/__tests__/coolify-client.test.ts
@@ -554,6 +554,28 @@ describe('CoolifyClient', () => {
       );
     });
 
+    it('should parse the deployments envelope Coolify returns for /deploy (#238)', async () => {
+      // Real Coolify returns { deployments: [{ message, resource_uuid, deployment_uuid }] },
+      // one entry per triggered deployment (a tag can match several apps).
+      mockFetch.mockResolvedValueOnce(
+        mockResponse({
+          deployments: [
+            { message: 'Deployment queued', resource_uuid: 'app-1', deployment_uuid: 'dep-1' },
+            { message: 'Deployment queued', resource_uuid: 'app-2', deployment_uuid: 'dep-2' },
+          ],
+        }),
+      );
+
+      const result = await client.deployByTagOrUuid('my-tag', true);
+
+      expect(result).toEqual({
+        deployments: [
+          { message: 'Deployment queued', resource_uuid: 'app-1', deployment_uuid: 'dep-1' },
+          { message: 'Deployment queued', resource_uuid: 'app-2', deployment_uuid: 'dep-2' },
+        ],
+      });
+    });
+
     it('should deploy by Coolify UUID (24 char alphanumeric)', async () => {
       mockFetch.mockResolvedValueOnce(mockResponse({ message: 'Deployed' }));
 
--- a/src/__tests__/mcp-server.test.ts
+++ b/src/__tests__/mcp-server.test.ts
@@ -1062,6 +1062,179 @@ describe('CoolifyMcpServer v2', () => {
       expect(deleteSpy).toHaveBeenCalledWith('svc-uuid', 'task-uuid');
     });
   });
+
+  describe('deploy tool handler', () => {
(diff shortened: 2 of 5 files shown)


StuMason:
Re the flagged uncertainty: confirmed by live use — the crunch estate's deploy-verification flow (2026-07-02) does exactly trigger-response `deployment_uuid` → `GET /deployments/{uuid}` and it resolves correctly. No smoke test needed for that assumption.

PR #272, opened 2026-07-14
fix: honor key filter on env_vars list so reveal exposes only the requested variable

### LINKED ISSUE #271: env_vars list ignores the key parameter — reveal=true dumps every secret on the resource
## What happens

The `env_vars` tool schema accepts `key` on every action, so an MCP client (LLM) that wants a single value naturally calls:

```json
{ "resource": "application", "action": "list", "uuid": "<uuid>", "key": "MY_FLAG", "reveal": true }
```

But the three `list` branches in `src/lib/mcp-server.ts` never use `key` — they always return the **full** env list. With `reveal: true` that means asking "what is MY_FLAG set to?" returns every variable on the resource in plaintext (API keys, DB passwords, PEM private keys) straight into the LLM context/transcript.

Observed in practice: an agent checking two boolean flags on an app with ~100 env vars received the complete plaintext dump three times into its transcript.

## Why it matters

The masking-by-default design (`***`) exists precisely to keep secrets out of MCP clients, and the tool description encourages `reveal=true` for the "what is FOO set to?" case — but that case is exactly the one where the caller only needs one value and instead gets all of them. The blast radius of a single innocuous read is the entire secret set of the resource.

## Expected

`action=list` + `key=FOO` returns only the `FOO` variable (masked, or plaintext with `reveal=true`). Without `key`, behaviour unchanged.

## Environment

`@masonator/coolify-mcp` 2.14.0 (also reproducible from current `main`, d81f281), Coolify v4 self-hosted.

I have a small PR ready with the fix + tests — will link it here.

artgas1:
Fixes #271

## Problem

The `env_vars` tool schema accepts `key` on every action, but the `list` branches ignored it and always returned the full env list. Combined with `reveal: true`, a caller asking "what is FLAG set to?" received every secret on the resource (API keys, DB passwords, private keys) in plaintext into the MCP client context.

## Fix

- `action=list` now filters the response to the requested `key` across all three resources (application / service / database). Filtering happens in the tool handler after the client fetch, so it applies to both masked and revealed projections.
- Tool description documents the `key` + `reveal` pattern ("always combine reveal with key so only the requested value is exposed").
- Behaviour without `key` is unchanged (full list, masked by default).

## Tests

Three new cases in `mcp-server.test.ts`:
- `list` + `key` returns only the matching variable (application)
- `list` + `key` + `reveal` — response contains the requested value and does **not** contain the other variable's plaintext (service)
- `list` without `key` still returns all variables (database)

`npm test` — 420 passed, `npm run lint` — 0 errors, prettier clean.

### DIFF
--- a/src/lib/mcp-server.ts
+++ b/src/lib/mcp-server.ts
@@ -1183,7 +1183,7 @@ export class CoolifyMcpServer extends McpServer {
     // =========================================================================
     this.tool(
       'env_vars',
-      "Manage env vars for app, service, or database. Values are masked by default (returned as '***') to avoid leaking secrets to MCP clients; pass reveal=true on the list action when the caller explicitly needs the plaintext (e.g. 'what is FOO set to?'). Set is_buildtime=false (and/or is_runtime=true) for runtime-only vars to avoid Dockerfile ARG issues with multiline values like PEM keys.",
+      "Manage env vars for app, service, or database. Values are masked by default (returned as '***') to avoid leaking secrets to MCP clients; pass reveal=true on the list action when the caller explicitly needs the plaintext (e.g. 'what is FOO set to?'). On list, pass key to return only that variable — always combine reveal with key so only the requested value (not every secret on the resource) is exposed. Set is_buildtime=false (and/or is_runtime=true) for runtime-only vars to avoid Dockerfile ARG issues with multiline values like PEM keys.",
       {
         resource: z.enum(['application', 'service', 'database']),
         action: z.enum(['list', 'create', 'update', 'delete', 'bulk_update']),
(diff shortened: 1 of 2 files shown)


src/lib/mcp-server.ts
```diff
@@ -1221,11 +1221,19 @@ export class CoolifyMcpServer extends McpServer {
         reveal,
         data,
       }) => {
+        // On `list`, an optional `key` narrows the response to that single
+        // variable. This matters most with reveal=true: without it, asking
+        // for one value dumps every secret on the resource to the MCP client.
+        const filterByKey = <T extends { key: string }>(vars: T[]): T[] =>
+          key ? vars.filter((v) => v.key === key) : vars;
```
StuMason, line 1228: 🥇

StuMason (approved):

PR #278, opened 2026-07-15
fix: mask database env var secrets by default (#274)

### LINKED ISSUE #274: Database env var listing returns secret values in plaintext by default
## Summary

The `env_vars` tool documents that secret values are **masked by default** (returned as `***`) so they aren't leaked to MCP clients, and that a caller must explicitly pass `reveal=true` to see plaintext. That promise holds for applications and services, but **not for databases**.

When you list the environment variables of a database, the values come back in **plaintext**, straight away, with no opt-in. Worse, passing `reveal=true` or `reveal=false` makes no difference to the database case at all — the flag is simply ignored.

## Steps to reproduce

1. Point the MCP server at a Coolify instance that has at least one database with env vars set (a password, a connection string, anything sensitive).
2. Call the `env_vars` tool with the database resource and the `list` action, leaving `reveal` unset (i.e. rely on the documented default).
3. Do the same for an application and for a service.

## Expected

All three resource types behave the same way the tool description promises: values masked as `***` by default, real values only when `reveal=true` is passed explicitly.

## Actual

- Application list → masked. ✅
- Service list → masked. ✅
- Database list → **plaintext secrets returned by default**, and `reveal` has no effect. ❌

## Why this matters

Database env vars are one of the most sensitive things this server can touch — they routinely hold DB credentials and connection strings. The masking layer exists specifically to stop those from being handed to an LLM client by accident. Right now the database path is a hole straight through it, and because the flag is ignored there's no way to even ask for masking. This also silently contradicts the tool's own documented default, so a caller trusting the docs is being misled.

StuMason:
## Summary

Closes #274.

The `env_vars` tool promises that secret values are **masked by default** (`***`) unless the caller passes `reveal=true`. Applications and services honored this; **databases did not**. `CoolifyClient.listDatabaseEnvVars` returned plaintext straight away, and the `env_vars` handler never forwarded `reveal` to it — so the flag was silently ignored and there was no way to even ask for masking.

Database env vars routinely hold DB passwords and connection strings, making this the highest-value leak of the three resource types, and it silently contradicted the tool's own documented default.

## Fix

- `listDatabaseEnvVars(uuid, { reveal })` now masks `value`/`real_value` by default via `maskEnvVar`, mirroring `listServiceEnvVars`. Plaintext only when `reveal: true`.
- The `env_vars` database `list` branch now forwards `reveal` to the client.

## Tests

- Client: masked-by-default and `reveal=true` returns plaintext.
- Handler: `reveal` is forwarded to `listDatabaseEnvVars`.
- Full suite: 422 passing, lint clean.

## Behaviour

| Resource | Before (default) | After (default) |
|---|---|---|
| Application list | masked ✅ | masked ✅ |
| Service list | masked ✅ | masked ✅ |
| Database list | **plaintext ❌** | masked ✅ |

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -7,6 +7,10 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ## [Unreleased]
 
+### Security
+
+- **`env_vars` list now masks database secrets by default** (#274) — the tool description promised every resource type masks values as `***` unless `reveal=true` is passed, but the database `list` branch called `listDatabaseEnvVars` with no options, so `CoolifyClient.listDatabaseEnvVars` returned **plaintext** straight away and the `reveal` flag was ignored entirely. Database env vars routinely hold DB passwords and connection strings, so this was the highest-value leak of the three. `listDatabaseEnvVars` now masks `value`/`real_value` by default (mirroring the application/service methods) and the `env_vars` handler forwards `reveal` to it, so databases finally match the documented default: masked unless the caller explicitly opts into plaintext.
+
 ## [2.14.1] - 2026-07-14
 
 ### Security
--- a/src/__tests__/coolify-client.test.ts
+++ b/src/__tests__/coolify-client.test.ts
@@ -4480,30 +4480,41 @@ describe('CoolifyClient', () => {
   // ===========================================================================
 
   describe('listDatabaseEnvVars', () => {
-    it('should list database env vars', async () => {
-      const mockEnvs = [
-        {
-          id: 1,
-          uuid: 'env-1',
-          key: 'DB_HOST',
-          value: 'localhost',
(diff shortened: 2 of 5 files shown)


StuMason:
Closing unmerged — superseded by #282, which is now on `main` as e8dfb4bf.

This PR was one of three parallel attempts at the same bug, filed as separate tickets in different languages (#274 English here, #275 French, #276 Russian) as a deliberate test. All three PRs produced a **byte-identical** production fix — the language of the ticket changed nothing about the code. #282 won purely on test depth: it was the only one to add a smoke test (which CLAUDE.md requires) and explicit `reveal=false` coverage.

Nothing wrong with the work here — it just lost a coin-toss decided by test coverage. Thanks for the review that ran on it; the fix it describes is the one that shipped.

Issue #302, opened 2026-07-29
v4.2 additive surface: destinations, app settings, spec re-vendor


StuMason:
Tracking the remaining v4.2 additions from #292 that are worth covering but did not justify their own issue. Split these out if any grows legs.

**Destination endpoints** ([coollabsio/coolify#10405](https://github.com/coollabsio/coolify/pull/10405)) — REST access to deployment destinations. The MCP already accepts `destination_uuid` when creating an application but gives no way to discover one, so today you have to get it from the UI. Listing destinations closes that loop, which makes this the highest-value item in this list despite being the smallest.

**Application settings on application endpoints** ([coollabsio/coolify#10188](https://github.com/coollabsio/coolify/pull/10188), #9017) — v4.2 exposes application settings and preview-deployment settings, plus custom Docker registry URLs. Some of this may already be reachable through `application update`; needs an audit of the v4.2 field list against the current `UpdateApplicationRequest` before deciding what is genuinely missing.

**Scheduled ClickHouse backups** — `database_backups` already covers scheduled backups generally. Needs checking whether ClickHouse works through the existing path or wants specific handling.

**Re-vendor `docs/coolify-openapi.yaml`** — the bundled spec is the ground truth for the drift check, and it predates v4.2. Refreshing it from upstream is a prerequisite for most of the above, since `check:spec-drift` will reject client routes the bundled spec has never heard of. Probably do this first. The external tracker referenced in #292 ([heavy-duty/cast#77](https://github.com/heavy-duty/cast/issues/77)) covers overlapping ground and is worth reading alongside.

**Per-team MCP server toggle** ([coollabsio/coolify#10771](https://github.com/coollabsio/coolify/pull/10771)) — not a gap, a positioning question. v4.2 ships its own built-in MCP server behind a per-team toggle. Worth understanding what surface that covers and how it overlaps with this project before planning V3 (#259).

Raised from the "new surface area" list in #292.

StuMason:
Sorted into its parts: the spec re-vendor is #347, destinations listing is #385, and the app-settings audit falls out of the re-vendor diff. The per-team MCP toggle was a positioning note and has been answered by the 3.x roadmap.

Issue #301, opened 2026-07-29
Service application and service database management (v4.2)


StuMason:
Coolify v4.2 added a service application management API ([coollabsio/coolify#9249](https://github.com/coollabsio/coolify/pull/9249)) plus service database management endpoints. The MCP covers neither — the client has no `/services/{uuid}/applications/...` or `/services/{uuid}/databases/...` routes at all.

These appear in the v4.2 breaking-change list as POST-only:

```
/services/{uuid}/applications/{app_uuid}/start|restart|stop
/services/{uuid}/databases/{database_uuid}/start|restart|stop
```

They were listed in #292 alongside the endpoints that broke, but they are **additive for us, not breaking**, because the client never called them. Confirmed by grep, so no action was needed in #296.

Why it is worth having: a Coolify service is a multi-container stack, and today the MCP can only act on the service as a whole. Restarting one misbehaving container inside a service means dropping to the UI. `control` already has the right shape for this — it just needs to reach one level deeper.

Note for implementation: these are POST-only and only exist from v4.2, so unlike the endpoints fixed in #296 there is no older GET form to fall back to. Pointing them at a pre-4.2 instance should give a clear "requires Coolify v4.2+" error rather than a bare 404.

Raised from the "new surface area" list in #292.

Issue #300, opened 2026-07-29
Database and service logs via API (v4.2)


StuMason:
Coolify v4.2 exposed database and service logs over the API, including per-container logs and timestamps ([coollabsio/coolify#6293](https://github.com/coollabsio/coolify/pull/6293)). The MCP has `application_logs` but no equivalent for databases or services.

This is the most immediately useful gap of the v4.2 additions. `diagnose_app` can pull an application's logs to explain why it is unhealthy; ask the same question about a database or a service and the server has nothing to look at. A service that will not start is exactly when you want logs, and right now that means the assistant sends you to the Coolify UI.

Worth covering:

- Database logs and service logs, reusing the existing `application_logs` shape
- Per-container selection, since a service is several containers and "the service logs" is ambiguous without it
- The same truncation and pagination discipline `application_logs` already applies (`lines`, `max_chars`) — log payloads are the easiest way to blow a context window

Also worth wiring into `diagnose_app` / `find_issues` once it exists, so an unhealthy service surfaces a log tail the way an unhealthy application does.

Raised from the "new surface area" list in #292.

Issue #299, opened 2026-07-29
Move resources between environments (v4.2 POST /move)


StuMason:
Coolify v4.2 added `POST /move` to relocate applications, databases and services between environments ([coollabsio/coolify#8968](https://github.com/coollabsio/coolify/pull/8968)). Not covered by the MCP.

This is a genuinely useful one for the assistant: "promote this app from staging to production" is a natural request that currently has no answer, and the manual alternative (recreate the resource in the target environment and copy env vars across) is exactly the kind of multi-step operation where a mistake is expensive.

Notes for whoever picks this up:

- It is state-changing, so it is `POST`-only and belongs to the group that has always accepted POST — no compatibility fallback needed (see the method-split note in CLAUDE.md).
- Only exists from v4.2, so it needs a clear error when pointed at an older instance rather than a bare 404.
- Should carry the destructive/blast-radius framing the delete actions have, since moving a resource between environments is not a read.

Raised from the "new surface area" list in #292.

PR #304, opened 2026-07-29
chore: re-vendor Coolify OpenAPI spec for v4.2 and generate the chunks

StuMason:
Prerequisite for the rest of the V2 line (#298–#301). Partially addresses #302 — the spec re-vendor half. The destinations, application-settings and ClickHouse-backup items on that issue remain open.

## Why this first

`docs/coolify-openapi.yaml` is the ground truth for "does Coolify support X", and `check:spec-drift` rejects any client route the bundled spec has never heard of. It predated v4.2, so every other v4.2 issue was blocked behind it.

## What changed upstream

**42 new paths, nothing removed** — purely additive, so this cannot break the drift check.

They map onto the open issues almost exactly:

| Issue | Paths that appeared |
| --- | --- |
| #298 tags | `/tags`, `/{applications,databases,services}/{uuid}/tags`, `.../tags/{tag_uuid}` |
| #299 move | `/{applications,databases,services}/{uuid}/move` |
| #300 logs | `/databases/{uuid}/logs`, `/services/{uuid}/logs`, `/services/{uuid}/applications/{app_uuid}/logs`, `/services/{uuid}/databases/{database_uuid}/logs` |
| #301 service app/db | `/services/{uuid}/applications/...`, `/services/{uuid}/databases/...` incl. start/restart/stop |
| #302 destinations | `/destinations`, `/destinations/{uuid}`, `/servers/{server_uuid}/destinations` |

Plus **two things nothing was tracking**, which I'll raise separately:

- **Volume backup schedules** — `/{applications,databases,services}/{uuid}/storages/{storage_uuid}/backups`, with `VolumeBackupScheduleRequest` / `VolumeBackupScheduleResponse` schemas
- **DigitalOcean and Vultr server provisioning** — `/servers/digitalocean`, `/servers/vultr` and their images/regions/sizes/plans/ssh-keys lookups, plus `/hetzner/firewalls` and `/hetzner/networks`. We already have a `hetzner` tool, so this is the same shape for two more providers.

Five new schemas: `VolumeBackupScheduleRequest`, `VolumeBackupScheduleResponse`, `ApplicationSetting`, `Destination`, `Tag`.

## The chunks are now generated

`docs/openapi-chunks/` is a developer reference — CLAUDE.md and the contributing guide both say "check the chunks first". It was hand-maintained, last touched 30 June, and would have gone stale the instant this PR landed. A stale reference that people are told to trust is worse than no reference.

`scripts/split-openapi-chunks.mjs` now generates it from the bundled spec, grouped by first path segment, and `npm run check:chunk-drift` fails CI if the committed chunks differ from freshly generated output. New `destinations-api`, `tags-api` and `cloud-providers-api` chunks; `untagged-api` still catches the bare instance routes.

The split is deliberately **line-based rather than parse-and-reserialise**. Round-tripping through a YAML library reformats the whole file and buries the real upstream diff in noise.

## Byte-fidelity with upstream restored

Prettier had reformatted the vendored spec at some point, so diffing it against a fresh upstream download showed whitespace churn rather than API changes — which defeats the point of vendoring it. Verified: the previous copy was **not** byte-identical to upstream; the new one is.

`docs/coolify-openapi.yaml` and `docs/openapi-chunks/` are now in a new `.prettierignore`, on the principle that vendored and generated artifacts shouldn't be hand-formatted. Re-vendoring is now a clean `curl > file`.

This was a genuine conflict between two CI steps: `prettier --check .` wanted to reformat the generated chunks, while `check:chunk-drift` compares them byte-for-byte. Ignoring beats accommodating.

## Docs fix

`site/concepts/how-it-works.md` described `search_docs` as "a local MiniSearch index over `docs/openapi-chunks/`". It actually fetches Coolify's published `llms-full.txt` at runtime and indexes that — it has never read the chunks. Corrected in the file tree and the architecture diagram. CLAUDE.md now says the chunks are generated and must not be hand-edited.

## Testing

- **458 tests pass** (445 before, 13 new in a new `scripts/__tests__/split-openapi-chunks.test.ts` suite).
- The interesting assertions are on the committed output, not just the pure functions: every spec path appears **exactly once** across the chunk set (nothing lost, invented or duplicated), every schema reaches `schemas.yaml`, every first path segment resolves to a real chunk file, and regenerating produces byte-identical output to what is committed.
- Verified independently via a YAML parse rather than the line regex: 138 spec paths in, 138 across the chunks, 18/18 schemas.
- `npm run check:spec-drift` OK (98 client routes), `check:chunk-drift` OK (14 chunks), lint 0 errors, build clean, `prettier --check` clean on a clean checkout.

No client or tool code changes — this is spec, tooling and docs only.


StuMason:
All six findings actioned — each verified against the code before acting rather than taken on trust.

**1. `main()` guard failing open — fixed, and you were right that this was the one that mattered.** Confirmed the sibling script already does it properly (`process.argv[1] === __filename` where `__filename = fileURLToPath(import.meta.url)`), so it now matches verbatim. The silent-pass failure mode was the real problem: `check:chunk-drift` exiting 0 having checked nothing is worse than it failing.

**2. `schemas.yaml` absorbing everything after `components:` — confirmed and fixed.** `grep -c '^tags:' docs/openapi-chunks/schemas.yaml` returned 1 before, at line 1130 exactly as you said, and `servers:` was dropped from all 14 chunks. Rather than just bounding the slice, `parseSpec` now sections the spec by column-0 keys and returns them explicitly. Unrecognised top-level keys **throw** — so if upstream appends `security:` or `webhooks:`, generation stops and someone decides, instead of it landing in schemas.yaml unnoticed. Verified: a synthetic spec with `webhooks:` throws.

**3. Hardcoded `HEADER` — fixed.** Now derived from the spec's own `openapi`/`info` sections. You put your finger on the actual inconsistency: a PR whose premise is "these are generated" had a curated header that would have asserted `3.1.0` and version `'0.1'` forever after an upstream bump.

**4. Bodies not asserted — fixed, and this was the sharpest catch.** You're right that key-level assertions pass in exactly the failure mode a line-based splitter has. Both are deep equality now: `expect(merged).toEqual(spec.paths)` and `expect(schemas.components?.schemas).toEqual(spec.components?.schemas)`.

**5. Tautological test — fixed.** It was restating the splitter's own constant. Now imports `SPEC_PATH` from `check-client-spec-drift.mjs` and compares the two modules, which is what the name always claimed.

**6. em-dash churn — reverted.** A `json.dumps` round-trip introduced it. `package.json`'s diff is now just the two new scripts.

Also added coverage for the deliberately-dropped keys (`servers`, `tags`) so a new upstream top-level key fails loudly.

**461 tests** on this branch (was 459), lint 0 errors, both drift checks green.

On the follow-up: confirmed `.github/coolify-openapi-baseline.yaml` genuinely does not exist, so the weekly drift job has never done anything — which is why 42 new v4.2 paths sat unnoticed until they were found by hand. Filed as #309, including your point that the vendored spec can now serve as the baseline directly. Same silent-pass shape as finding 1, which is a nice thread between the two.

Issue #303, opened 2026-07-29
V3: OAuth 2.1 + Streamable HTTP transport — run as a container inside Coolify


StuMason:
Part of the V3 epic (#259). This is the change that makes V3 a major rather than an internals refresh.

## Goal

Ship one codebase with **two transports**:

1. **stdio** — exactly what exists today. Unchanged behaviour, unchanged config, unchanged npm and MCPB artifacts. Nobody on stdio should notice V3 happened beyond a version number.
2. **Streamable HTTP + OAuth 2.1** — deployable as a container on the user's own Coolify instance, so remote MCP clients (claude.ai in particular) can manage infrastructure without anything running locally.

The tool layer is already transport-agnostic: `mcp-server.ts` holds all 42 tools and `index.ts` only wires stdio. An HTTP entry point is additive, not a rewrite.

## Why OAuth 2.1 rather than a shared token

Remote MCP clients require it, but it is also straightforwardly better than the alternative:

- The long-lived Coolify API token stays **server-side as container config**. It never travels to the MCP client.
- The client only ever holds a **short-lived, audience-scoped, revocable** access token.
- A client can be revoked without rotating anything in Coolify.
- Sessions are attributable, so tool calls can be audited.

Compared with the status quo — a static Coolify token pasted into a client config file in plaintext, valid forever, unattributable — this is a clear improvement.

It does increase attack surface: an internet-facing service that brokers control of an estate. That is a deliberate, opt-in trade with the same trust model as exposing Coolify's own API, and the README should say so plainly rather than bury it.

## Auth model: env-var Coolify token, OAuth2 authenticates the human

Coolify cannot be the authorization server — upstream authenticates its API with Sanctum bearer tokens, has no OAuth2 endpoints, and this project is staying independent. So the container implements the AS itself.

**The Coolify API token is container configuration, supplied as an environment variable**, exactly as the stdio server takes `COOLIFY_ACCESS_TOKEN` today. Setting env vars is the native idiom when deploying anything on Coolify, so this needs no extra UI and no new concepts.

**OAuth 2.1's job here is authenticating the human to the container**, not brokering Coolify credentials. The flow proves who you are, then issues a short-lived, audience-scoped, revocable access token. The Coolify token never leaves the container.

This gets the full security story with almost none of the machinery:

- The long-lived Coolify token stays server-side and is never transmitted to any client. That property comes from OAuth2 itself, not from where the token was sourced.
[...]

Issue #259, opened 2026-07-11
V3 Epic: SDK v2 + OAuth 2.1 remote transport


StuMason:
## What V3 is

Two things, and the second is the one that matters:

1. **Migrate to MCP TypeScript SDK v2** and spec revision 2026-07-28.
2. **Add an OAuth 2.1 + Streamable HTTP transport** (#303) so the server can run as a container on a user's own Coolify instance and be driven by remote MCP clients like claude.ai — while stdio keeps working exactly as it does today.

Worth being honest about the first on its own: swapping the SDK, keeping every tool name and schema, means a stdio user notices nothing but a version number. That is not a major. **#303 is what makes this a 3.0** — new config surface, new deployment topology, new artifact, new security posture.

## Scope

- [ ] Swap `@modelcontextprotocol/sdk` → `@modelcontextprotocol/server`
- [ ] `registerTool()` migration — done ahead of time in #260, which works on SDK 1.x and carries over unchanged
- [ ] Verify stdio under the stateless core against Claude Desktop / Claude Code / Cursor
- [ ] **OAuth 2.1 + Streamable HTTP transport (#303)**
- [ ] Elicitation on the v2 `inputRequired` API (#261) — optional on stdio, **mandatory in HTTP mode**, see #303
- [ ] Structured tool output (#262), paired with #264 so the schemas earn their token cost
- [ ] Tasks extension for long-running deploys (#263)
- [ ] MCP Apps dashboard extension (#264)
- [ ] Re-cut the MCPB bundle + `server.json` against v2
- [ ] Container image + Coolify service template (#303)
- [ ] Bump major: npm `3.0.0`

## What is NOT in V3

Everything that runs on SDK 1.x ships on the V2 line first, labelled `v2`:

- #302 re-vendor the OpenAPI spec — prerequisite for all the Coolify-surface work, since `check:spec-drift` rejects routes the bundled spec has never heard of
- #300 database and service logs
- #260 tool annotations + `registerTool` — zero token cost, gets read-only tools parallel-dispatched today, and is free V3 prep
- #298 tags, #299 `/move`, #301 service application and database management

Doing these first means V3 starts from a codebase that already covers v4.2's API surface, rather than trying to land API coverage and a transport rewrite in the same release.

## What changes in SDK v2 / spec 2026-07-28

[...]

PR #310, opened 2026-07-29
feat: annotate all 43 tools and migrate off the deprecated tool() API

### LINKED ISSUE #260: Tool annotations (readOnlyHint/destructiveHint) + registerTool migration — works on SDK 1.x today
## Why (do this BEFORE V3 — works on the current SDK)

None of our 42 tools declare [tool annotations](https://modelcontextprotocol.io/specification/2025-03-26/changelog) (`readOnlyHint`, `destructiveHint`, `idempotentHint`, `openWorldHint`) — spec-stable since 2025-03-26 and supported by SDK 1.x `registerTool()`. Client behavior is real now:

- **Claude Code parallel-dispatches tools marked `readOnlyHint: true`** — our ~15 read-only tools (`list_*`, `get_*`, `diagnose_*`, `search_docs`, …) currently serialize when they could fan out
- **VS Code Copilot** shows confirmation dialogs for anything NOT read-only — unannotated read tools get scary prompts they don't deserve
- **ChatGPT dev mode** badges READ/WRITE from annotations; Claude Desktop groups tool UI by them

For a server whose tool list includes `deploy`, `stop_all_apps`, `bulk_env_update`, and four flavors of `delete`, the safety signal of `destructiveHint` is exactly what the spec authors had in mind.

## What

1. Migrate `mcp-server.ts` from legacy `this.tool(name, desc, schema, handler)` to `this.registerTool(name, { description, inputSchema, annotations }, handler)` — mechanical sweep, and it future-proofs for SDK v2 (#259) which keeps the `registerTool` shape.
2. Annotate all 42 tools. Suggested mapping:
   - `readOnlyHint: true` — `get_version`, `get_mcp_version`, `get_infrastructure_overview`, `list_*` (5), `get_*` (4), `server_resources`, `server_domains`, `diagnose_app`, `diagnose_server`, `find_issues`, `search_docs`, `deployment` (get/list actions)
   - `destructiveHint: true` — `application`/`database`/`service`/`projects`/`environments`/`private_keys`/`github_apps`/`storages`/`scheduled_tasks`/`database_backups` (delete actions), `stop_all_apps`, `bulk_env_update`, `control` (stop/restart), `deploy`, `redeploy_project`, `restart_project_apps`, `system` (disable_api)
   - `idempotentHint: true` where PATCH semantics apply (updates)
   - `openWorldHint: true` on everything (remote Coolify API)
   - Mixed-action tools take worst-case annotations (a tool with any destructive action is `destructiveHint: true`, `readOnlyHint: false`)
3. Tests: assert annotations on registered tools (snapshot of name → annotations map).

## Notes

- Zero token cost — annotations ride the existing tools/list response.
- Also consider Claude Code's proprietary `anthropic/maxResultSizeChars` meta-annotation for `application_logs` / `deployment get`.

Refs: #259 (V3 epic). Annotation support survey: https://mcpblog.dev/blog/2026-03-13-mcp-tool-annotations

StuMason:
Closes #260. Not stacked — branches off `main` directly.

## What

All 43 tools now declare `readOnlyHint` / `destructiveHint` / `idempotentHint` / `openWorldHint`, and registration moves off the SDK's **deprecated** `tool()` overloads to `registerTool()`.

Client behaviour this changes today, on SDK 1.x, with no upgrade:

- Claude Code parallel-dispatches tools marked read-only — 21 of ours currently serialise when they could fan out
- VS Code Copilot raises confirmation dialogs for anything not marked read-only, so reads stop getting prompts they don't deserve
- ChatGPT dev mode badges READ/WRITE from these

`registerTool` is also the shape SDK v2 keeps, so this is #259 prep that pays off immediately.

## Classification

**21 read-only**, **20 destructive**, **2 neither** — each checked against the tool's actual action enum, not its name.

The two that aren't obvious:

- **`hetzner`** — provisions paid servers, which is consequential, but every action is additive: it creates, it never replaces or removes. `destructiveHint` means "may perform destructive updates", not "may cost money", so it's `false`. Flagging it since it's the one I'd expect pushback on.
- **`validate_server`** — a write (`POST`), but re-running converges on the same state, so `idempotentHint: true` and not destructive. No reason for a client to gate it behind confirmation.

Consolidated tools take worst-case values: `env_vars` is destructive because it has a `delete`, even though most of its actions aren't.

## One table, not 43 call sites

Annotations live in a single `TOOL_ANNOTATIONS` map, and all 43 registrations go through one `defineTool` wrapper that looks them up. A tool missing from the table **throws at construction** rather than silently shipping unannotated.

The point of annotations is that the safety posture of the surface is auditable — that's much easier to review as one table than as 43 scattered literals, and it makes the V3 `readonly` mode (#303) a one-line filter over this map rather than a hand-maintained list that drifts.

## Correction to the issue

#260 says annotations are "zero token cost — they ride the existing tools/list response". They ride it, but they are **not free**. Measured:

| | tokens added to `tools/list` |
| --- | --- |
| Every hint spelled out | **~751** |
| Non-default hints only | **~415** |

Against a ~6,600 token budget that's 11% versus 6%. Per spec the defaults are `readOnlyHint=false`, `destructiveHint=true`, `idempotentHint=false`, `openWorldHint=true`, so emitting those tells a compliant client nothing it didn't already assume.

`destructiveHint: true` is the deliberate exception — it *is* the default, but it's also the safety signal, and stating it explicitly is worth the bytes. There's a test asserting the payload stays under 2000 bytes so the defaults can't creep back in.

## Testing

- **488 tests pass** (478 before, 10 new). Coverage 98.3% statements / 99.55% functions.
- Every registered tool has annotations; the table and the registered set are asserted equal **in both directions** (a table entry with no tool is dead config; a tool with no entry can't register).
- No tool is both read-only and destructive.
- The read-only and destructive sets are asserted by explicit name list, so a future tool silently landing in the wrong bucket fails.
- `get_mcp_version` is the only `openWorldHint: false` — it returns a local constant while everything else reaches the Coolify API.
- `defineTool` throws for an unknown tool name.
- The token-budget guard above.

Verified against the built server that all 43 tools actually carry annotations through to registration, rather than trusting the source.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -15,8 +15,6 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
   **Correction to the issue's premise:** #260 said annotations were zero-cost because they ride the existing `tools/list` response. They ride it, but they are not free — spelling out every hint measured at ~751 tokens against a ~6,600 token budget. Only non-default hints are emitted now (per spec: `readOnlyHint=false`, `destructiveHint=true`, `idempotentHint=false`, `openWorldHint=true`), which halves it to ~415 tokens, with a test that fails if the defaults creep back in.
 
-### Added
-
 - **`logs` tool: container logs for databases and services, not just applications** (#300) — `diagnose_app` could explain an unhealthy application by reading its logs, but the same question about a database or a service had no answer and sent you to the Coolify UI, which is exactly the moment you want logs.
 
   Added as **one consolidated `logs` tool** with a `resource` parameter (`application` / `database` / `service`), matching the `env_vars` and `control` pattern, rather than three or four sibling tools — the token budget is the point of the v2.0.0 design. `application_logs` still works and is marked superseded in its description; it will be removed in v3. Also adds `show_timestamps`, which upstream has always supported and this client never sent.
@@ -27,6 +25,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 - **Log endpoints returned an object behind a `string` type** (#300) — `getApplicationLogs` was declared `Promise<string>` but handed back Coolify's `{ logs: "..." }` envelope unchanged, so any caller doing string work on it would have failed at runtime. Confirmed against a live Coolify instance; the unit tests had mocked a bare string, which is why it went unnoticed. Responses are now unwrapped to a plain string, with bare strings still accepted for older instances.
 
+- **Corrected the documented behaviour of `search_docs`** — `site/concepts/how-it-works.md` described it as "a local MiniSearch index over `docs/openapi-chunks/`". It actually fetches Coolify's published `llms-full.txt` at runtime and indexes that; it has never read the chunks.
+
 ### Changed
 
 - **Re-vendored `docs/coolify-openapi.yaml` from upstream, now covering Coolify v4.2** (#302) — the bundled spec is ground truth for "does Coolify support X" and predated v4.2, so it was missing **42 paths** (nothing was removed). The additions line up with the open v4.2 issues: tags on applications/databases/services (#298), `/move` between environments (#299), database, service and per-container logs (#300), service application and database management (#301), and destinations (#302). Also new and not previously tracked: volume backup schedules on storages, and DigitalOcean/Vultr server provisioning alongside Hetzner firewalls and networks. Five new schemas: `VolumeBackupScheduleRequest`, `VolumeBackupScheduleResponse`, `ApplicationSetting`, `Destination`, `Tag`.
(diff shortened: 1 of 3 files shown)


StuMason:
All six done, plus both test gaps you flagged as non-blocking.

**1. Duplicate `### Added`** — fixed, and chasing it turned up a **duplicate `### Fixed`** as well (from the #302 `search_docs` entry) that neither of us had spotted. `[Unreleased]` is now one section per type: Added / Fixed / Changed.

**2. Stale JSDoc** — you were right about the cause, it was written from the issue text before I measured. Replaced with the measured numbers, since as you say that sentence was the one most likely to talk the next person into re-adding the defaults.

**3. Threshold now scales.** Your arithmetic checks out — ~1,660 bytes against a 2,000 ceiling is ~8 tools of headroom, and the failure message would have blamed the defaults for what was really just growth. Now asserts a per-tool average under 50 bytes; all four hints spelled out is ~90, so it still catches the actual regression.

**4. Compile-time enforcement** — done via `satisfies` + a `ToolName` key union. Verified with a negative type check that an unknown name now fails `tsc`:

```
error TS2322: Type '"tool_that_does_not_exist"' is not assignable to type
'"application" | "database" | ... 31 more ... | "validate_server"'
```

Runtime throw kept as the backstop, and the existing test still reaches it through the `unknown` cast unchanged.

**5. Frozen** — both constants. Twenty tools sharing one mutable reference that `registerTool` then stores was a genuinely nasty failure mode for how cheap the fix is.

**6. Moved** below `docsSearch`.

**Both test gaps taken too.** Annotations are now asserted over a real `tools/list` round trip via `InMemoryTransport`, so the tests check the wire format rather than the registry — and the byte-budget test measures the actual payload instead of my reconstruction of it, which makes finding 3's number trustworthy rather than arithmetic I did once.

**On your consolidation point** — that's the sharpest thing in the review and I've put it in the table comment rather than just fixing the code, because it's a trap for #303 specifically: `env_vars` list, `deployment` get/list_for_app and `system` health/list_resources are pure reads sitting under destructive tools, so a read-only mode filtering on this map would drop them, which is not what someone asking for read-only access expects. Gate on action, not tool. Noted in the comment so it's found in the right place.

Also taking the `validate_server` wrinkle — good to know the first call installs Docker, so it's not as inert as "idempotent" implies even though the classification holds. And agreed on `hetzner`: `readOnlyHint: false` is what clients gate on, and leaving `idempotentHint` at its `false` default carries the real signal, that two calls make two servers.

**489 tests** (was 488), lint 0 errors, prettier clean, both drift checks green.

Left the shared test-helper suggestion for #259 — you're right that `_registeredTools` is now more call sites for that upgrade, but it's a refactor of existing convention rather than something this PR introduced.

PR #313, opened 2026-07-29
feat: elicit human confirmation for destructive operations, and fix the project batch tools (#261)

### LINKED ISSUE #261: Elicitation: interactive confirmation for destructive operations (capability-gated)
## Why

The server can pause and ask the human before executing destructive operations — `stop_all_apps`, `application delete`, `database delete` (with volumes!), `bulk_env_update` across N apps. [Elicitation](https://modelcontextprotocol.io/specification/2025-06-18/changelog) is spec-stable since 2025-06-18 and works on SDK 1.x.

Today `stop_all_apps` uses a `confirm` boolean param, which the *model* fills in — that's the model confirming with itself. Elicitation puts the confirmation in front of the *human*, in client UI.

## Client support (the constraint)

- **Claude Code**: supported since v2.1.76 (2026-03-14), incl. Elicitation hooks
- **VS Code Copilot**: supported
- **Claude Desktop / claude.ai**: lagging (open issue anthropics/claude-ai-mcp#153)
- **Cursor**: unclear

So: **strictly progressive enhancement.** Gate on the client's `elicitation` capability at runtime; when absent, fall back to the current `confirm`-param behavior unchanged.

## What

1. Helper: `confirmDestructive(server, message): Promise<boolean>` — checks client capability, elicits if available, otherwise returns true (defer to existing param-based guards).
2. Wire into: `stop_all_apps`, all `delete` actions (application/database/service/project/environment — especially `delete_volumes: true`), `redeploy_project`, `bulk_env_update` when touching >3 apps.
3. Message should carry blast radius: "Stop ALL 12 running applications on 3 servers?" — we have the counts from the pre-flight lists.
4. Tests: mock client capabilities both ways.

## Note for V3

SDK v2 redesigns this as `inputRequired.elicit()` / `InputRequiredResult` (#259) — keep the helper thin so only its internals change.

Refs: #259. Claude Code support: https://claudelab.net/en/articles/claude-code/mcp-elicitation-support

StuMason:
Closes #261.

## The problem

`stop_all_apps` is gated like this:

```ts
{ confirm: z.literal(true) }
```

That parameter is filled in by the **model**. It is the model confirming with itself before taking every application on the estate down. The same gap applies to the deletes, which have no gate at all beyond the client's own annotation-driven prompt.

[Elicitation](https://modelcontextprotocol.io/specification/2025-06-18/changelog) moves the question to the human, rendered by the client, outside the model's control.

## What it does

Confirmation is requested for `stop_all_apps`, `redeploy_project`, the application / database / service / project / environment deletes, and `bulk_env_update` when it touches more than three apps. Under that threshold it stays silent on purpose — a prompt people have learned to click through protects nothing.

Prompts state their blast radius rather than asking in the abstract:

```text
EMERGENCY STOP: take down 12 running applications
(billing-api, checkout-web, mailer-worker, search-indexer and 8 more) across 3 servers?

Every one stays down until it is started again. This is estate-wide, not scoped to a project.
```

```text
Delete application "checkout-web" (a1b2c3d4e5f6)?

Persistent volumes will be DESTROYED and their data is not recoverable
(delete_volumes was not set, and it defaults to true). This cannot be undone.
```

That volume sentence is the sharp edge. `delete_volumes` is documented `default: true` on all three DELETE endpoints, so **omitting the optional flag destroys the data** — the opposite of what "optional boolean, left unset" reads like at a call site. Only an explicit `false` is reported as "volumes kept", which is both what the spec says and the safe way to be wrong if the spec is lying again.

Env var *values* never appear in a prompt. `bulk_env_update` names the key and the app count, because the prompt surfaces in client UI and in logs, and echoing the value would leak whatever secret is being rotated.

## Progressive enhancement, not a new requirement

Client support is uneven — Claude Code and VS Code Copilot have it, Claude Desktop and claude.ai do not yet. So the server checks the client's advertised `elicitation` capability at runtime and, when it is absent, approves and lets the existing parameter guards stand. **A client that cannot be asked is not a client that gets blocked.**

Once a client *does* advertise support it fails closed in the other direction: decline, cancel, timeout and transport errors all abort. The one failure that does not abort is the blast-radius lookup, because a summary we could not compute is a reason to ask a vaguer question, not to skip asking — the human gets asked anyway, and told the lookup failed.

The timeout is five minutes rather than the SDK's 60-second default, which is a fine ceiling for a machine answering and a bad one for a person reading "stop ALL 12 applications?" and deciding.

The lookups that produce the counts are lazy, so clients that will never show the question do not pay for the extra API call.

## Two bugs caught by running it live

Both were found against a real Coolify 4.1.2 over stdio, not by re-reading the spec — the same lesson as the 2.15.0 regression.

1. **`GET /applications` does not populate `server_uuid`.** It nests `destination.server_id` instead. The first version of the stop-all prompt counted servers by `server_uuid`, which made `across N servers` dead code that could never once have fired. `Application.destination` is now in the types, documented as verified live.

2. **`.filter(Boolean)` on `server_id` discards Coolify's localhost server**, which is id `0`. Every application on the test estate reports it. The count would have said "1 server" for an estate split across localhost and one remote — undercounting the blast radius, the one direction a confirmation must never be wrong in.

The live check itself is side-effect free by construction: the test client answers `decline`, and the guard only calls the underlying operation on `accept`, so the only requests reaching Coolify are the read-only pre-flight lookups.

## Also in here

`isRunningStatus` is extracted from `stopAllApps` and shared with the prompt. Two copies of "which apps count as running" would eventually disagree, and the failure mode of that is a confirmation dialog understating its own blast radius.

Behaviour is unchanged, **including a known over-match**: `'unhealthy'` contains `'healthy'`, so `exited:unhealthy` is classified as running. That has always been true of `stop_all_apps`; extracting the predicate did not introduce it and this PR does not alter it. It costs a no-op stop against an already-stopped app and errs toward over-stating the blast radius, which is the safe direction. Flagging rather than quietly changing it — happy to fix in a follow-up if you'd rather.

## Tests

552 passing (+28). `elicit.ts` is at 100% statements, branches, functions and lines.

The tests drive a real `Client` over an in-memory transport rather than stubbing `getClientCapabilities()`, because the thing most likely to be wrong here is the capability negotiation itself: a server that elicits against a client which never advertised support fails at runtime in exactly the environments (Claude Desktop, claude.ai) where nobody is watching a test suite. Declaring the capability on the client and letting the SDK negotiate is the only way the "absent" branch is genuinely absent.

One branch is unreachable over a real transport — the SDK always wraps a rejection into an `McpError`, so `elicitInput` rejecting with a non-Error needs a stub. That test says so rather than pretending the wire-level one covers it.

## V3

`confirmDestructive` is deliberately thin. SDK v2 redesigns this as `inputRequired.elicit()` (#259), and everything version-specific is inside that one function — call sites only see `ConfirmOutcome`. It is also a stated prerequisite for the HTTP transport in #303, where the parameter-based guards stop being credible at all.

## Tool count

Unchanged at 44.


StuMason:
Addressed. Findings 1 and 2 both turned out to be real, and 2 was worse than it looked.

## 1 — abort signal: fixed, with a test that fails without it

`extra.signal` is now threaded through `guardDestructive` → `confirmDestructive` → `elicitInput`.

The regression test reproduces the exact sequence you described, scaled down: the client times the `tools/call` out at 250ms, and the human accepts at 600ms. I checked it actually fails with the signal removed rather than passing vacuously — my first version had the human never answer at all, which passes either way and proves nothing.

On the timeout tension: with the signal threaded, the client's cancellation is authoritative and `ELICIT_TIMEOUT_MS` becomes a backstop for clients that never cancel, so I've left it at five minutes. Added a README note that a client may need its tool timeout raised, and that the server aborts cleanly when it doesn't — you see a failed call, not something running behind your back.

## 2 — confirmed, and it is a shipped no-op bug

You were right to push on this. `GET /applications` does not return `project_uuid` at all: **0 of 26 applications** on the live instance carry it, and it is absent from the response entirely.

So `redeployProjectApps` and `restartProjectApps` have both been filtering on a field that never arrives. Both matched zero applications on every call and returned a cheerful `{"summary":{"total":0,"succeeded":0,"failed":0}}` while doing nothing. That is not a #261 bug — it has been shipped behaviour — but the prompt would have said "Redeploy 0 applications in this project?", which is how it surfaced.

The only link an application carries is the numeric `environment_id`. Resolution is now project → `GET /projects/{uuid}` → `environments[].id` → applications in those environments, in a shared `applicationsInProject` helper. Verified live: six projects that all previously reported zero now report their real contents.

```
[1] Redeploy 1 application (…) in this project?
[3] Redeploy 3 applications (…, …, …) in this project?
```

Your instinct that "`redeployProjectApps` has the same dependency so it probably *is* populated" was the reasonable inference — and it was wrong, which is rather the point of the PR.

## 3 — TOCTOU and the duplicated filter: done

Took your suggestion. The client methods now accept the resolved set, the guard passes through exactly what the human was shown, and the duplicated project filter is gone into `applicationsInProject`. Removes the double fetch too.

## 4 — no-op prompts: done

`summarize` may return `null` for "nothing to confirm". An emergency stop on an idle estate and a redeploy of an empty project both run without asking.

## 5 — unescaped names: done

`sanitizeForPrompt` flattens C0/C1 control characters and clamps to 64 chars, applied to resource names, env var keys, and blast-radius lists.

## 6 — added two, filed the rest

Added `restart_project_apps` (agreed, it read as an oversight sitting next to `redeploy_project`) and `system disable_api` — you're right that it's the most self-locking call in the server, and it's the one Stu personally asked about when this work started.

The remaining unguarded deletes are #315, framed as "where does the line sit before prompt fatigue makes every dialog worthless" rather than "which are destructive". The `isRunningStatus` over-match is #314, as you asked.

## Smaller notes

- Zero-field elicitation rendering: I've driven the full round trip over real stdio against a live instance, so the protocol path is verified, but I have not eyeballed how Claude Code and Copilot actually draw it. Flagging that as genuinely unverified rather than claiming otherwise.
- On the coverage claim you couldn't check: 567 passing now, `elicit.ts` at 100% statements/branches/functions/lines.

New tests cover cancellation-during-prompt, both no-op paths, the sanitiser, `disable_api` guarded and `enable_api` deliberately not, and that each batch operation receives the exact approved set.

StuMason:
Round two addressed. Findings 1–3 were all real; 1 is the one I should have caught myself.

## 1 — model-supplied uuids: fixed

You're right that this is the stronger vector, and it's a bad miss given `elicit.ts` spends a paragraph reasoning about the *weaker* one. The names are chosen by someone with write access to the instance; the uuids are chosen by a model that only needed to read a hostile README. I sanitized the source I'd thought about and left the source I hadn't.

All three sites now sanitize the identifier. A real 36-char uuid is well inside the 64-char clamp so genuine values render unchanged, and there's a test pinning that alongside one per injection site.

## 2 — `applicationsInProject` reintroducing the silent zero: fixed

This is the sharper of the two. `project.environments ?? []` converts "I could not resolve this project" into "this project has no applications", which is the same cheerful `0 succeeded` the method exists to eliminate, arriving from a different direction. And your point about the #261 path is the part that would have hurt: `summarize()` returns `null`, which auto-approves, so an unresolvable project produces *neither* a dialog nor an error nor any work.

Now throws when `environments` is absent. An empty array is left alone — that's a real answer about a real project, and the distinction is exactly the one worth drawing. Test added for the absent case, which was the gap you identified.

## 3 — mixed key spaces: fixed

Keys are prefixed by source (`id:` / `uuid:`). It doesn't merge the two spaces — nothing could without a second lookup — but one server described both ways no longer silently becomes two, and the split is deliberate rather than accidental. Test covers the mixed response.

## 4 — stateless HTTP: noted where it will be read

Added to the `supportsElicitation` doc comment rather than a changelog line nobody greps. Fail-open is right on stdio, where the alternative is blocking Claude Desktop out of tools that work today; it's very probably wrong for a remote server, which is the transport where the parameter guards stop being credible at all. Flagged as a decision to make deliberately in #303 rather than inherit.

## 5 — remaining unguarded ops

Leaving these to #315, where I've framed it as "where does the line sit before prompt fatigue makes every dialog worthless" rather than "which are destructive". `private_keys delete` is listed first there, and I take your point that it's the strongest candidate — but pulling it forward means also deciding `cloud_tokens delete`, which is the identical shape, and that's the slope #315 exists to handle in one go.

## On the bundling

Fair, and the PR title and body now say so. It genuinely was found while building the prompt, and the prompt is wrong without it, but "elicitation PR" was the wrong label for something that also fixes two tools that have never worked.

## Smaller

- `restartProjectApps(uuid, [])` — pinned. An approved-nothing that silently becomes a restart-everything after a `|| await …` refactor is a good thing to have a test for.
- `isRunningStatus` — your split-on-`:` suggestion is in #314, including the point that `running:unhealthy` must stay *in* scope.

574 tests, `elicit.ts` still 100% on all four axes. Re-verified live afterwards: the six projects still resolve correctly, so the new throw isn't firing on real responses.

PR #320, opened 2026-07-30
feat: confirmation for the credential deletes (#315)

### LINKED ISSUE #315: Elicitation: decide the boundary for the remaining unguarded destructive tools
Follow-up to #261 / #313, so the covered set is a decision rather than a gap.

#313 wired human confirmation into the operations with the largest blast radius: `stop_all_apps`, `redeploy_project`, `restart_project_apps`, `system disable_api`, the application / database / service / project / environment deletes, and `bulk_env_update` above three apps.

Still unguarded, in descending order of how much it probably matters:

- `private_keys` `delete` — removing the key an application deploys with breaks future deploys, and the key material is not recoverable from Coolify.
- `cloud_tokens` `delete` — same shape; the token cannot be read back.
- `github_apps` `delete` — breaks every application sourced from that installation.
- `database_backups` `delete` / `delete_execution` — deleting a backup schedule is quiet and only hurts later, which is arguably an argument *for* a prompt.
- `storages` `delete` — persistent storage definitions.
- `scheduled_tasks` `delete`.
- `deployment` `cancel` — recoverable by redeploying, so likely not worth it.
- `control` `stop` / `restart` on a single resource — probably correctly left alone; single-resource lifecycle is the routine case and prompting on it is how people learn to dismiss prompts.

## The actual question

Not "which of these are destructive" — the annotations already answer that — but **where the line sits before prompt fatigue makes every dialog worthless**. The `bulk_env_update` threshold exists for exactly this reason. A confirmation people have learned to click through protects nothing, and adding eight more prompts is the fastest way to get there.

My instinct is that the irreversible-credential deletes (`private_keys`, `cloud_tokens`) are worth it and the rest are not, but that is a judgement call worth making deliberately.

Whatever is decided, the covered set should stay enumerated explicitly in the README and `site/tools/index.md` rather than implied, as it is now.

StuMason:
Closes #315.

Draws the boundary the issue asked for, and records it so it is a decision rather than an accident: a delete gets a prompt when the loss is **irrecoverable** or **estate-wide**. Everything else stays silent, because a dialog on every delete is how dialogs stop being read.

## Guarded

- **`private_keys delete`** — key material is write-only in Coolify; once deleted it only comes back if you still hold the original.
- **`cloud_tokens delete`** — same shape; the token value cannot be read back.
- **`github_apps delete`** — estate-wide: every application sourced from the installation loses its deploy source at once, and the prompt counts them:

  ```text
  Delete GitHub app "example-installation" (id 1)?

  19 applications (billing-api, checkout-web, mailer-worker and 16 more)
  sourced from it will lose their deploy source and cannot deploy until re-linked.
  ```

  The count filters by `source_type` as well as `source_id`, because the numeric id can collide with a GitLab source. Verified live: `source_type` is the Laravel class name (`App\Models\GithubApp`), null for public-repo apps.

## Deliberately not guarded

Storages, scheduled tasks, individual env vars, backup schedules, `deployment cancel`, single-resource `control` stop/restart. A test pins the storages case, so if one of these ever grows a guard it is a deliberate revisit of the boundary, not a side effect.

## Verified

- 598 unit tests (+5), `elicit.ts` still 100% on all four coverage axes
- Live over stdio against a real 4.1.2 instance with a declining client: private-key and GitHub-app prompts render with real names and a real 19-application count; decline aborts; nothing deleted

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -9,6 +9,8 @@ and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0
 
 ### Added
 
+- **`private_keys update` with new key material is guarded like the delete** (#315 review) — overwriting key material IS deleting the old key: Coolify never returns key material, so the previous value is exactly as gone either way, and a model "fixing" a key by overwriting it takes the same servers offline. Renames and description edits pass without a prompt.
+
 - **Confirmation for the credential deletes** (#315) — `private_keys delete`, `cloud_tokens delete` and `github_apps delete` now elicit like the other destructive operations. The boundary drawn, recorded here so it is a decision rather than an accident: a delete gets a prompt when the loss is **irrecoverable** (key material and token values are write-only in Coolify — once deleted, they only come back if you still hold the original) or **estate-wide** (deleting a GitHub app breaks every application sourced from it, and the prompt counts them: "2 applications (api, worker) sourced from it will lose their deploy source"). Routine deletes — storages, scheduled tasks, individual env vars, backup schedules, `deployment cancel` — deliberately stay unprompted, and a test pins that so growing a guard is a revisit of the boundary, not a side effect. Prompt fatigue is the failure mode: a dialog on every delete is how dialogs stop being read.
 
   The GitHub-app blast radius filters by `source_type` as well as `source_id`, because the numeric id can collide with a GitLab source. Verified live: `source_type` is the Laravel class name (`App\Models\GithubApp`), null for public-repo applications.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,14 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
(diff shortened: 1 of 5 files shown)


StuMason:
Addressed. Finding 1 was the important one and the fix is exactly your line.

**1 — fixed.** Unknown `source_type` now counts. Your framing was right: the field is absent from the vendored spec and only live-verified on 4.1.2, so absence is plausible, and excluding on it printed the most reassuring sentence the dialog can emit on precisely the delete the guard exists for. Public-repo apps still drop out on their null `source_id`, so the change only bites when a numeric id genuinely matches — where over-counting asks harder, the module's chosen direction. Test added for the absent-type shape.

On your related question: checked live — `github_apps list` returns only the two real installations; the seeded public source is not addressable through this surface, and public-repo apps carry null for both source fields. So the "delete the public source, break every public app, get told nothing breaks" scenario has no route in through this tool today.

**2 — guarded.** You're right that replacing key material *is* deleting the old key, and "a model fixing a key by overwriting it" is a realistic path. `update` with `private_key` set now elicits with replacement wording; renames and description edits pass silently. Tests for both.

**3 — fixed** ("every guarded tool").

**4 — declined, with evidence.** Checked live before deciding: `GET /servers` does **not** return `private_key_uuid` on 4.1.2 — the field exists in the type and not in the response, which is the exact `server_uuid`/`project_uuid` pattern this repo keeps finding. Counting servers would need N per-server fetches for a prompt, and the generic wording is honest. If a future version serialises it, cheap to revisit.

**5 — fixed.** The zero-app prompt now names its consequence: not undoable from Coolify, re-linking means re-creating the app on GitHub.

**Tests:** the degraded two-call path is covered (`listApplications` rejects → prompt still fires carrying the label), plus the summary-projection comment so nobody "optimises" `listApplications()` into zeroing the count. 603 total.

PR #321, opened 2026-07-30
test(site): contact endpoint coverage + astro check in CI (#319)

### LINKED ISSUE #319: Site: tests and type-checking for the contact endpoint
Follow-up to #318, from its review. Non-blocking — the site is not deployed yet.

**1. No tests on `site/src/pages/api/contact.ts`.** It has ~200 lines of validation, sanitisation, origin checking and rate limiting, and the site has no test runner. `clean()`, `EMAIL_RE`, `originAllowed()` and `rateLimited()` are pure or near-pure and the handler takes a synthetic `Request`. The CRLF stripping in `clean()` is the header-injection defence for the SES subject and reply-to — that should not be regression-testable only by hand. Vitest plus one `contact.test.ts` covers it.

**2. Nothing type-checks the site.** `site/tsconfig.json` extends `astro/tsconfigs/strict` but no step runs `astro check`, so the strict config is decorative. Add `@astrojs/check` + `typescript` and a step in `.github/workflows/site.yml`. It would catch things like the `querySelector(...) as HTMLButtonElement` cast in `index.astro`, which casts away a possible null and then dereferences it.

**3. Rate limiting is per-process.** Fine on a single instance, which is what runs today. If the site is ever scaled to more than one replica the limit becomes per-replica and needs shared state — noted in the code comment.

## Before deploying

- `AWS_ACCESS_KEY_ID`, `AWS_SECRET_ACCESS_KEY`, `AWS_REGION`, `CONTACT_FROM` (SES-verified identity) on the Coolify app
- Confirm the SES account is out of sandbox, or it only delivers to verified recipients
- Coolify app needs its build changed: it currently expects static nginx on port 80; the new image is Node on 4321

StuMason:
Closes #319.

## What

The site had no test runner — `prettier` was the entire devDependency list — while `src/pages/api/contact.ts` takes unauthenticated POSTs and feeds form fields into an SES **subject** and **reply-to**. The CRLF stripping in `clean()` is the header-injection defence, and nothing regression-tested it.

- **21 vitest cases** drive the real `POST` handler with synthetic `Request`s, SES mocked at the module boundary. Header injection (name → subject, email → reply-to), validation, honeypot, origin checking (same-origin, `x-forwarded-host` behind Traefik, garbage origin, missing origin), rate limiting (per-IP window, first-hop of a multi-hop `x-forwarded-for`), and both failure-honesty paths (SES down → 502 naming an alternative route; credentials absent → 503 rather than pretending to send).
- **`astro check` in CI** — `tsconfig` extends `astro/tsconfigs/strict` and nothing ran the compiler, so the strictness was decorative.
- **The `querySelector(...) as HTMLButtonElement` cast is fixed properly** — `querySelector<T>` keeps the null, and a missing submit button now no-ops instead of throwing mid-submit. (A legal cast, so `astro check` alone would never have flagged it.)

## Test design notes

Everything goes through the real handler rather than exporting the private helpers — the property that matters is end-to-end, and the wiring between helpers is where an escape would live. Each request carries a distinct forwarded IP because the rate limiter is module-level state; without that, test order changes would produce spurious 429s.

Item 3 from the issue (per-process rate limiting) is left as the code comment already covering it — single replica today, and shared state is a deploy-shape decision, not a test gap.

## Not covered here

The pre-deploy checklist from the issue (SES env vars on the Coolify app, sandbox status, build config change from static nginx to Node on 4321) is deployment work, not repo work — it still stands.

### DIFF
--- a/.github/workflows/site.yml
+++ b/.github/workflows/site.yml
@@ -47,6 +47,18 @@ jobs:
       - name: Format check
         run: npm run format:check
 
+      - name: Type check
+        # tsconfig extends astro/tsconfigs/strict; without this step that
+        # strictness is decorative — nothing else runs the compiler over
+        # .astro files.
+        run: npm run check
+
+      - name: Test
+        # The contact endpoint takes unauthenticated POSTs and feeds form
+        # fields into SES subject/reply-to headers. Its CRLF stripping is the
+        # header-injection defence, and this is what regression-tests it.
+        run: npm test
+
       - name: Build
         # Also exercises the tool count read from src/lib/mcp-server.ts, which
         # throws if the registration shape changes.
--- a/site/package-lock.json
+++ b/site/package-lock.json
@@ -14,8 +14,70 @@
         "astro": "^5.16.2"
       },
       "devDependencies": {
+        "@astrojs/check": "^0.9.10",
         "prettier": "^3.9.6",
-        "prettier-plugin-astro": "^0.14.1"
+        "prettier-plugin-astro": "^0.14.1",
+        "typescript": "^5.9.3",
+        "vitest": "^4.1.10"
+      }
+    },
+    "node_modules/@astrojs/check": {
+      "version": "0.9.10",
+      "resolved": "https://registry.npmjs.org/@astrojs/check/-/check-0.9.10.tgz",
+      "integrity": "sha512-zgx/UQMozdjOa3bOxjgeCFdtpE3c9rRX6xHwa+2QXvy8z8Akifu2AtubHyv/zzC2znO8dl8fFWL4K+Ba9kS8HQ==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "@astrojs/language-server": "^2.16.7",
+        "chokidar": "^4.0.3",
+        "kleur": "^4.1.5",
+        "yargs": "^18.0.0"
+      },
+      "bin": {
+        "astro-check": "bin/astro-check.js"
+      },
+      "peerDependencies": {
+        "typescript": "^5.0.0 || ^6.0.0"
+      }
+    },
+    "node_modules/@astrojs/check/node_modules/chokidar": {
+      "version": "4.0.3",
+      "resolved": "https://registry.npmjs.org/chokidar/-/chokidar-4.0.3.tgz",
+      "integrity": "sha512-Qgzu8kfBvo+cA4962jnP1KkS6Dop5NS6g7R5LFYJr4b8Ub94PPQXUksCw9PvXoeXPRRddRNC5C1JQUR2SMGtnA==",
+      "dev": true,
+      "license": "MIT",
+      "dependencies": {
+        "readdirp": "^4.0.1"
+      },
+      "engines": {
+        "node": ">= 14.16.0"
+      },
+      "funding": {
+        "url": "https://paulmillr.com/funding/"
+      }
+    },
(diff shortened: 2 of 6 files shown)


StuMason:
All nine addressed; 21 → 29 cases.

**2 (taking it first — the real gap):** `company` is now tested exactly like `name`. You were right that the headline property was half-covered — it is the second attacker-controlled field landing in the subject.

**1:** The multi-hop test now carries the trust assumption in a comment: these tests document *bucketing*, not a security guarantee — first-hop XFF is client-supplied unless the edge rewrites it, and if inbound XFF is ever trusted, only `GLOBAL_CEILING` stands. Which is also now tested (60 distinct IPs → 61st gets 429), so the backstop you called the real cliff is pinned.

**3 + 4:** Took the fake-timers suggestion wholesale — `beforeEach` advances past `WINDOW_MS`, so every test starts with an empty map and the global budget coupling is gone entirely, not managed. That unlocked everything at once: window expiry (the map's only eviction path), the `x-real-ip` fallback, and the `unknown` single-bucket collapse — with the test comment naming it as the failure most likely to actually happen given the deploy is mid-move from nginx to Node.

**5:** Exact assertion — the subject must equal `coolify-mcp enquiry — ` + 120 a's, so the test fails when the limit moves.

**6:** Comment corrected to say the opposite and true thing: the guard reads env at *request* time, which is precisely why the 503 test can delete a var after import. Both credential vars now tested via `it.each`.

**7:** `as never` replaced with a `makeContext` helper returning a typed `APIContext` — narrow on purpose, with a comment saying why: if the handler starts reading more of the context, tests should fail loudly on the missing property.

**8:** One multipart case — the encoding the real browser form sends.

**9:** Warm-up 200s asserted in the multi-hop test, `From`/`Destination` pinned in the happy path (neither may ever be wired from form input), `formData()`-throw → 400 covered, and `@types/node` declared explicitly.

The client-script coverage gap you noted stands — out of scope here, and the `getElementById` cast you cleared is left as-is.

StuMason:
Both pre-merge items fixed, and 2 was a real product bug — thank you for insisting the suite decide the contract rather than inherit it.

**1:** Both tests now set `host` explicitly, with a comment explaining why (`new Request()` never populates Host; undici adds it at dispatch). The cross-origin 403 now comes from the comparison and the garbage-origin 403 from the parse `catch` — the mutations you named no longer survive.

**2:** Fixed in the endpoint, not just pinned. `clean()` is now header-bound only (name, email, company); the message goes through `cleanBody()`, which normalises CRLF to LF and clamps but keeps newlines — the flattening was a header defence applied to something that is not a header. A two-paragraph test pins the corrected contract, including that `\r` still never reaches the delivered mail.

**3:** All four covered — message clamp pinned at exactly 5000, blank company omits both the suffix and the `Company:` line, the honeypot `console.warn` is asserted with its shape (your point about untested log lines being what refactors delete silently is exactly right), and the 503 body now carries the same `hey@stumason.dev` assertion as the 502.

**4:** `Buffer.byteLength`, the guarded env restore, the honest makeContext comment, and a chunked-hole comment on the 413 test — all in. Left the file-scope fake timers as-is while this is the only test file; agreed on the pairing when a second lands.

**5:** Confirmed against branch protection: required checks are `build (20.x/22.x/24.x)` only — the Site job is not required, so the pending-"Expected" trap does not apply.

33 cases, `astro check` clean, build green.

PR #322, opened 2026-08-03
feat: add service sub-application management (FQDN, image, start/stop/restart)

### LINKED ISSUE #301: Service application and service database management (v4.2)

fedesenmartin:
## ⚠️ Requirements

**Coolify v4.2.0 or later is required** for this feature. The `PATCH /services/{uuid}/applications/{app_uuid}` endpoint was added in v4.2.0.

To upgrade Coolify:
```bash
curl -fsSL https://cdn.coollabs.io/coolify/install.sh | bash -s 4.2.0
```

---

## Summary

Adds support for updating service sub-applications via the Coolify v4.2 API endpoint `PATCH /services/{uuid}/applications/{app_uuid}`.

## Context

Coolify services are multi-container stacks where each container is a sub-application with its own FQDN, Docker image, and settings. Previously, the MCP server only supported updating the entire service (via `docker_compose_raw`), but not individual sub-applications.

This fills the gap described in Issue #301 — there was no way to update a sub-application's URL (FQDN), image, or other properties without modifying the entire service compose file.

### 1. TypeScript Types (`src/types/coolify.ts`)
- Added `UpdateServiceApplicationRequest` interface for the PATCH request body
- Added `ServiceApplication` interface for the response

### 2. Client Method (`src/lib/coolify-client.ts`)
- Added `updateServiceApplication(serviceUuid, appUuid, data)` method
- Follows existing patterns: uses `cleanRequestData()` and returns typed response
- Placed logically after other service methods

### 3. MCP Tool Action (`src/lib/mcp-server.ts`)
- Added `update_application` action to the `service` tool
- Parameters:
  - `uuid` (service UUID, required)
  - `app_uuid` (sub-application UUID, required)
  - `url` (optional FQDN string)
  - `image` (optional Docker image string)
  - `human_name` (optional display name)
  - `description` (optional)
  - `exclude_from_status` (optional boolean)
  - `is_gzip_enabled` (optional boolean)
  - `is_stripprefix_enabled` (optional boolean)
- Updated tool description to document the new action

## Implementation Notes

- Follows existing code patterns and conventions
- Uses the same error handling and validation approach as other service actions
- The endpoint accepts partial updates — only provided fields are sent to Coolify
- No destructive operation guard needed (this is an update, not a delete)

## Testing

The implementation follows the established patterns in the codebase:
- Client method mirrors other PATCH operations (e.g., `updateService`, `updateApplication`)
- Tool action follows the same structure as other service actions
- Type definitions match the Coolify OpenAPI spec for this endpoint

## Example Usage

```typescript
// Update a sub-application's FQDN
await client.updateServiceApplication(serviceUuid, appUuid, {
  url: "https://app.example.com"
});

// Update Docker image
await client.updateServiceApplication(serviceUuid, appUuid, {
  image: "nginx:latest"
});

// Multiple fields at once
await client.updateServiceApplication(serviceUuid, appUuid, {
  human_name: "My App",
  description: "Production web server",
  is_gzip_enabled: true
});
```

Closes #301

---

## Example AI Prompts

> "Change the FQDN of the Ghost Blog sub-application in my ghost service to https://newdomain.example.com"

This will trigger the `update_application` action on the `service` tool, calling:
```
service action=update_application uuid=<service-uuid> app_uuid=<app-uuid> url="https://newdomain.example.com"
```

Other example prompts:
- "Update the Ghost blog docker image to ghost:5"
- "Rename the Ghost Blog sub-application to 'My Tech Blog'"
- "Enable gzip compression for the Ghost sub-application"

### 4. Version Compatibility Error Handling
- Added error hint in `errorHint()` function to detect 404 on `/services/{uuid}/applications/{app_uuid}` path
- Returns clear message: "Sub-application management requires Coolify v4.2+. Check version with `get_version`. Upgrade: `curl -fsSL https://cdn.coollabs.io/coolify/install.sh | bash -s 4.2.0`"
- Prevents confusing bare 404 errors when running against Coolify < v4.2.0
- Added test coverage for the new error hint

### DIFF
--- a/src/__tests__/coolify-client.test.ts
+++ b/src/__tests__/coolify-client.test.ts
@@ -6017,6 +6017,18 @@ describe('errorHint', () => {
     expect(errorHint(404, '/applications/app-uuid/tags')).toMatch(/different resource type/);
   });
 
+  it('hints at the v4.2 requirement for a 404 on a service application route', () => {
+    // Sub-application management (PATCH /services/{uuid}/applications/{app_uuid})
+    // was added in v4.2.0. Without this hint, the generic uuid mismatch fires.
+    expect(errorHint(404, '/services/svc-uuid/applications/app-uuid')).toMatch(/v4\.2/);
+    expect(
+      errorHint(404, '/services/svc-uuid/applications/app-uuid?force_domain_override=true'),
+    ).toMatch(/v4\.2/);
+    expect(errorHint(404, '/services/svc-uuid/applications/app-uuid')).toMatch(
+      /update_application/,
+    );
+  });
+
   it('hints at a possible resource-type mismatch for a 404 on a uuid route', () => {
     expect(errorHint(404, '/applications/app-uuid')).toMatch(/different resource type/);
   });
--- a/src/__tests__/coolify-client.test.ts
+++ b/src/__tests__/coolify-client.test.ts
@@ -2665,6 +2665,90 @@ describe('CoolifyClient', () => {
         expect.objectContaining({ method: 'POST' }),
       );
     });
+
+    it('should update a service sub-application', async () => {
+      mockFetch.mockResolvedValueOnce(mockResponse(mockApplication));
+
+      const result = await client.updateServiceApplication('svc-uuid', 'app-uuid', {
+        url: 'https://example.com',
+        human_name: 'My App',
+      });
+
+      expect(result).toEqual(mockApplication);
+      expect(mockFetch).toHaveBeenCalledWith(
+        'http://localhost:3000/api/v1/services/svc-uuid/applications/app-uuid',
+        expect.objectContaining({ method: 'PATCH' }),
+      );
+      const callBody = JSON.parse(mockFetch.mock.calls[0][1]?.body as string);
+      expect(callBody.url).toBe('https://example.com');
+      expect(callBody.human_name).toBe('My App');
+    });
+
+    it('should update a service sub-application with force_domain_override', async () => {
+      mockFetch.mockResolvedValueOnce(mockResponse(mockApplication));
+
+      await client.updateServiceApplication(
+        'svc-uuid',
+        'app-uuid',
+        { url: 'https://example.com' },
+        { forceDomainOverride: true },
+      );
+
+      expect(mockFetch).toHaveBeenCalledWith(
+        'http://localhost:3000/api/v1/services/svc-uuid/applications/app-uuid?force_domain_override=true',
+        expect.objectContaining({ method: 'PATCH' }),
+      );
+    });
+
+    it('should start a service sub-application', async () => {
+      mockFetch.mockResolvedValueOnce(mockResponse({ message: 'Started' }));
+
+      const result = await client.startServiceApplication('svc-uuid', 'app-uuid');
+
+      expect(result).toEqual({ message: 'Started' });
+      expect(mockFetch).toHaveBeenCalledWith(
+        'http://localhost:3000/api/v1/services/svc-uuid/applications/app-uuid/start',
+        expect.objectContaining({ method: 'POST' }),
+      );
+    });
+
+    it('should start a service sub-application with force and latest', async () => {
+      mockFetch.mockResolvedValueOnce(mockResponse({ message: 'Started' }));
+
+      await client.startServiceApplication('svc-uuid', 'app-uuid', {
+        force: true,
+        latest: true,
+      });
+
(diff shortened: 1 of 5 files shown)


StuMason (changes requested):
Thanks for this! It lands right on #301, which had been sitting parked, and you clearly did the homework: the endpoint, the `force_domain_override` query param and the request body all match the 4.2 OpenAPI spec exactly. The v4.2 hint on the 404 is exactly the house pattern, and its test passes (604/604 with it). Great to see.

First, an apology about the red X: the CI failure on your branch was mostly ours. A fresh npm advisory broke the audit step on main before your code ever got built, and that's fixed on main now (#324), so your next push gets an honest run.

That honest run will still fail though, and here's why. The branch doesn't compile: the handler reads five fields the tool schema never declares, so `npm run build` fails with six errors. Sneaky part: `npm test` stays green regardless, because ts-jest runs with `isolatedModules` and skips type-checking, so the phantom fields just read as `undefined`. Easy to miss locally, the build is the only thing that catches it.

```
src/lib/mcp-server.ts(1502): Property 'human_name' does not exist on type 'ShapeOutput<...>'
  (same for image, exclude_from_status, is_gzip_enabled, is_stripprefix_enabled)
src/lib/mcp-server.ts(1510): Argument of type 'string | undefined' is not assignable to 'string'
```

Changes requested:

1. Declare the missing fields in the `service` tool schema: `human_name`, `image`, `exclude_from_status`, `is_gzip_enabled`, `is_stripprefix_enabled`. Worth exposing `is_log_drain_enabled` at the same time, since your `UpdateServiceApplicationRequest` already carries it.
2. The error at line 1510: TypeScript won't narrow `args.app_uuid` through that guard, so pull it into a local after the early return (`const appUuid = args.app_uuid;`) and pass that.
3. The `url` description says "Null to clear", but `z.string().optional()` can't carry a null through the tool schema. Either drop that sentence or make the field nullable if clearing is a case you want to support.
4. The PR title promises start/stop/restart, but the diff doesn't include them. Happy either way: trim the title, or add them. The endpoints are in the spec and they'd follow the same shape.
5. Tests for the new client method and the tool action, please. `coolify-client.test.ts` has plenty of PATCH examples to crib from, and `mcp-server.test.ts` shows the tool-action pattern.

`npm run build && npm test` before pushing and you're golden.

One thing I'll pick up on our side after this lands: `force_domain_override: true` proceeds despite domain conflicts, which can pull a live domain off another resource, so it wants the same human-confirmation guard the destructive tools got in 2.17. Don't worry about that here unless you fancy it.

Keen to get this in. Thanks again for taking it on!

fedesenmartin:
Thanks for the review! All feedback addressed:

- Added missing schema fields (human_name, image, exclude_from_status, is_gzip_enabled, is_stripprefix_enabled, is_log_drain_enabled)
- Fixed TypeScript narrowing issue
- Removed "Null to clear" from url description
- Added start_application, stop_application, restart_application actions
- Added tests for new client methods and tool actions
- Implemented force_domain_override guard using guardDestructive pattern

All 626 tests pass, build and lint clean. Branch rebased onto latest main.

Ready for re-review!

StuMason (approved):
Re-review done: all five requested changes landed plus the optional extras (lifecycle actions and the force_domain_override guard). Verified locally before CI: build clean, 626 tests pass, schema passes the provider sweep test. Thanks for sticking with it through two rounds.

PR #325, opened 2026-08-04
fix: stop_all_apps schema breaks every Google Gemini request

phxmg:
## Summary

`stop_all_apps` declares `confirm: z.literal(true)`. zod emits that as `const: true`, and `@ai-sdk/google` rewrites `const` into `enum: [const]` when it converts the tool list for `generateContent`. Google only accepts string enum values, so every request that includes this server's tools fails:

```
400 Invalid value at 'tools[0].function_declarations[42].parameters.properties[0].value.enum[0]' (TYPE_STRING), true
```

Google rejects the whole request, not just the one declaration, so all 44 tools go down with it, and nothing in the error points at a schema. Anthropic and the OpenAI-compatible providers accept the same list unchanged, which is why this only bites Gemini users. Reproduced with opencode 1.17.19 + gemini-2.5-flash on published v2.18.0, and independently with curl against the live API.

## Changes

- `confirm` is now `z.boolean()` with the requirement in its description, and the handler's existing check (now `confirm !== true`) is the gate.
- Nothing is loosened. The literal never stopped a model from sending `true` (with the fixed schema Gemini still emits `confirm: true` unprompted), and the elicitation prompt from #261 is untouched. Non-booleans (`"true"`, `1`, omitted) are still refused by the parser; `false` is refused before any Coolify call is made.
- Added a test that walks every tool's `inputSchema` over a real `tools/list` round trip and fails on any non-string `enum` or `const`, so the next literal gets caught in CI instead of by a provider.
- Updated the `elicit.ts` docstring that quoted the old declaration.

## Test plan

- [x] Tests added/updated — 610 pass (+7; the new ones fail on `main`)
- [x] Manually tested locally
  - The same 44-tool payload captured out of `@ai-sdk/google`: 400 before, 200 after against live `gemini-2.5-flash`
  - `confirm: false` / omitted / `"true"` / `1` / `null` all refuse over stdio JSON-RPC; with a request-logging server on the other end, `confirm: false` produces zero HTTP requests
  - `confirm: true` against a mock hits exactly `GET /applications` then `POST /applications/{uuid}/stop`, running apps only
  - Read-only tools verified against a live Coolify 4.1.2
  - opencode + `gemini-2.5-flash`: 400 on published v2.18.0, clean `get_version` on this branch

## Checklist

- [x] Documentation updated if needed (CHANGELOG; tool count unchanged so no README edit)

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,16 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **Every Google Gemini request failed while this server was connected.** `stop_all_apps` declared `confirm: z.literal(true)`, which zod emits as `const: true`; `@ai-sdk/google` rewrites a JSON Schema `const` into `enum: [const]` when it converts the tool list, and Google's `enum` accepts strings only. `generateContent` answered `400 Invalid value at 'tools[0].function_declarations[42].parameters.properties[0].value.enum[0]' (TYPE_STRING), true` — and it rejects the **request**, not the offending declaration, so all 44 tools went down with the one. The symptom is a client that cannot make a single Gemini call until it disables this tool by name, which does not look like a schema problem from the outside. Anthropic and the OpenAI-compatible providers accepted the same list unchanged, which is why it survived from #261 to here.
+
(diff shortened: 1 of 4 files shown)


StuMason:
This is a brilliant bug report and a better fix. You did not just swap the schema, you proved the whole chain (zod emitting const, the ai-sdk rewrite to enum, Google only taking strings) and left a tripwire test so the next literal gets caught in CI instead of by a provider. I reproduced all of it locally before merging: the new sweep test fails on main and passes here, and the human confirmation gate is untouched.

Genuinely grateful, this had been silently breaking every Gemini user of the server. Merging now and cutting a patch release today so they are unblocked.

StuMason (approved):
Verified the whole chain locally: zod emits the literal as const true, ai-sdk rewrites const to enum, Google rejects non-string enum values. New sweep test fails on main and passes here, confirmation gate untouched, 610 tests green.

PR #345, opened 2026-08-11
feat: verify exact application environment

ninelhodzic:
Adds a read-only verify_app_environment tool for callers that already hold exact application, project, and environment anchors.\n\nThe implementation uses only GET /applications/{application_uuid} and GET /projects/{project_uuid}/{expected_environment}; it does not enumerate applications, projects, environments, or resources. Application UUID, project UUID when returned, environment name, and numeric environment identity are checked fail-closed. The response is limited to identity and name.\n\nAlso updates the tool annotation, contract snapshot and roster, tool-count manifests, README, CLAUDE guidance, and changelog.\n\nValidation:\n- npm run build\n- npm test -- --runInBand (661/661)\n- npm run lint (0 errors; 52 existing warnings)\n- npm run format:check\n- npm run check:chunk-drift (14/14)\n- npm --prefix evals run snapshots (4/4)\n- npm --prefix evals run typecheck\n- npm --prefix evals run lint\n- npm --prefix evals run format:check\n- git diff --check\n\nNo provider-backed integration test was run because this change is intentionally provider-free and read-only.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Added
+
+- **Exact application-environment verification without enumeration.** The new read-only `verify_app_environment` tool requires an application UUID, project UUID and expected environment name, reads only those exact application and environment endpoints, and returns a minimal identity proof. UUID, project, environment name and numeric environment identity drift all fail closed; embedded environment resources and configuration never enter the tool response.
+
 ## [2.19.3] - 2026-08-06
 
 A security release, and the one that ends the leak class instead of patching another instance of it. Update from any earlier version.
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -4,7 +4,7 @@ This file provides guidance to Claude Code (claude.ai/code) when working with co
 
 ## Project Overview
 
-MCP (Model Context Protocol) server for Coolify that provides 44 token-optimized tools for AI assistants to manage infrastructure through natural language. Tools cover servers, projects, environments, applications, databases, services, deployments, private keys, teams, cloud tokens, documentation search, smart diagnostics, and batch operations. v2.0.0 reduced token usage by 85% (from ~43,000 to ~6,600 tokens) by consolidating related operations into single tools with action parameters.
+MCP (Model Context Protocol) server for Coolify that provides 45 token-optimized tools for AI assistants to manage infrastructure through natural language. Tools cover servers, projects, environments, applications, databases, services, deployments, private keys, teams, cloud tokens, documentation search, smart diagnostics, and batch operations. v2.0.0 reduced token usage by 85% (from ~43,000 to ~6,600 tokens) by consolidating related operations into single tools with action parameters.
 
 ## Commands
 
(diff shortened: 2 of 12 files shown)


StuMason:
Thanks for this, and welcome. It is careful work: the fail-closed checks are right, the response projection keeps embedded resources out of the tool result, and the tests cover the drift cases properly. I pulled the branch and it is green locally (661/661, build, lint).

Two things before it can land.

**1. The CI red is not yours.** `build (20/22/24)` fail on `npm audit --audit-level=high` because a js-yaml advisory (GHSA-5p4m-2wfm-xmqj) landed after main last ran. Main fails the same way. I will fix that on main separately; once it is in, a rebase will turn your build green. `claude-review` failing is a known no-op, ignore it.

**2. I would rather this was an action than a 45th tool.** Every top-level tool costs tokens on every request, and the whole pitch of this server is consolidating operations behind action params. `environments` already takes `project_uuid` and `name`, so the same behaviour fits as:

```
environments { action: 'verify_app', project_uuid, name, application_uuid }
```

Same client method, same checks, same tests, and no README / CLAUDE.md / manifest / server.json count churn. Two smaller asks in the same pass:

- Return something a caller can use directly, e.g. `{ verified: true, application_uuid, environment: { id, uuid, name } }`. The environment uuid is not sensitive, and stripping it just forces the next call.
- Drop the one-off path-segment regex. Nothing else in the client validates segments that way, and a single method doing it differently is worse than either everyone or no one. Happy to take a separate PR that hoists a check into `request()` if you want that.

One question so I understand the shape: what is the calling flow in your agent that needs the proof in a single call, rather than `get_application` plus `environments get` and comparing `environment_id`? If it is a pre-mutation guard I get it, and it might inform where the action sits.

ninelhodzic:
Thanks — agreed on all of this.

The calling flow is ReleaseRun v2 inventory qualification. It starts with exact application, project, and environment anchors and uses the result in an immutable read-only observation receipt. That receipt gates a later, separately approved local inventory-row update; this is not a Coolify mutation guard.

The reason for one call is custody, not atomicity or saving a round trip: the connector owns the two reads, comparison, and projection, so only the bounded result crosses into the agent context. We do not want the agent composing two generic tool responses and becoming the implementation of that check.

`environments { action: "verify_app", project_uuid, name, application_uuid }` preserves that boundary and is the better public shape. I will move it there, use the richer response, drop the segment regex, and remove the tool-count churn.

I also see #348 has landed, so I will rebase the revision onto current `main`. Thanks for the careful review.

StuMason:
Quick update from our side: we're cutting the 3.0 release now, and we'd still like this capability in — the `environments { action: 'verify_app' }` shape we agreed on is a good fit for the 3.x line.

Since it's been a few weeks, here's an offer rather than a chase: if ReleaseRun work has you busy, we're happy to do the rework ourselves on top of your branch — your commits stay, you get the credit, and we'd just add the fold-in on top before merging. If you'd rather finish it yourself, that's great too and there's no deadline pressure from the 3.0 cut (it would land in a 3.x minor either way).

If we haven't heard back in a week or so we'll assume the first option is fine and pick it up. Thanks again for the careful work on this one.

StuMason:
Landed in #381 for 3.1 with your commit intact and credited (`a886f064`, authored by you), reworked to the shape we agreed here: `environments { action: 'verify_app' }` rather than a 46th tool, returning `{ verified, application_uuid, environment: { id, uuid, name } }`, with the path-segment regex dropped. Your fail-closed drift checks and no-list-calls guarantee are unchanged, and your tests carried over. You'll see it in the 3.1.0 notes — thank you for the careful work. Closing this in favour of #381.

PR #349, opened 2026-08-18
fix: #336 field-test follow-ups — db collision merge, env get default, preview twins, exact-match resolution, find_issues signals

### LINKED ISSUE #336: Field-test follow-ups: list_databases drops types on id collision, env schema, diagnose_app scope + FQDN matching, find_issues signals
Remaining items from the external field test of the remote server (the masking findings shipped in 2.19.3 / #334).

1. **`list_databases` silently drops whole database types (upstream id collision).** Verified live: an instance with 2 standalone Postgres and 2 standalone Dragonfly returns only the Dragonflys from `GET /databases`. Coolify keeps a per-type id sequence (both pairs are ids 1 and 2), and the list endpoint evidently merges the per-type collections keyed on `id`, so one type overwrites another. `/resources` reports all four correctly. Fix: augment `list_databases` from `/resources` merged by uuid; also repairs `get_infrastructure_overview` database counts.
2. **`environments get` schema/handler mismatch**: schema marks only `action` + `project_uuid` required, handler rejects with `Error: name required`. Align, and default to the sole environment when the project has exactly one.
3. **`diagnose_app` env-var count reads as a bug**: Coolify auto-creates a preview twin for every production application env var and the API returns both scopes merged, so counts double. Surface `is_preview` / count distinct keys. Never dedupe the underlying rows: deleting a twin destroys preview config.
4. **`diagnose_app` name resolution**: an exact FQDN match should win before substring matching produces a multi-app disambiguation error.
5. **`find_issues` ignores signals it already has**: an app in `running:unknown` and an available proxy patch upgrade produced zero findings. Fold known status/version signals in.
6. **(shipped)** version string on the v3 branch now reports `3.0.0-alpha.1` instead of the v2 version.

kashik0i:
Closes #336 — items 1–5, in the issue's order (item 6 already shipped on the v3 branch).

## 1. `list_databases` no longer silently drops whole database types

Coolify keeps a per-type id sequence and `GET /databases` merges the per-type collections keyed on that id, so two types sharing ids shadow each other. `listDatabases` now fetches `/resources` alongside and merges back in, by uuid, any row `/databases` dropped — every `standalone-*` resource type is a database. The merge is a no-op when `/databases` is complete; if `/resources` fails the plain `/databases` result is returned unchanged. Both `get_infrastructure_overview` counts and `find_issues` database coverage go through `listDatabases`, so the same merge repairs them. The merged rows pass through the central `deepSanitize` boundary like every other response, so nothing unmasked leaks in.

## 2. `environments get` honours its own schema

`name` was always optional in the schema; the handler rejected anyway. It now defaults to the sole environment when the project has exactly one; with several the error names them (`name required — project has 2 environments: production, staging`), and an environment-less project says so. Tool description updated, contract snapshot reviewed and updated to match.

## 3. `diagnose_app` env counts stop reading as a bug

The response now carries `distinct_keys`, `production_count`, `preview_count` and a per-variable `is_preview` alongside the unchanged raw `count`, so the preview-twin doubling is legible instead of alarming. No row is deduped, per the issue's warning — deleting a twin destroys preview config.

## 4. Exact matches win name resolution

`resolveApplicationUuid` runs an exact pass (name, or any entry of the comma-separated `fqdn` with scheme and trailing slashes ignored) before substring matching, so `api.example.com` resolves even when another app's FQDN contains it. Same for `resolveServerUuid` — IPs are substrings of each other constantly (`10.0.0.5` is inside `10.0.0.50`). Multiple exact matches still disambiguate.

## 5. `find_issues` folds in the signals it already had

Issues gain `severity: 'critical' | 'warning'`. Resources in `running:unknown` warn (running container, no working healthcheck, a failure would go undetected). Servers with an available Traefik update warn with the version pair — `traefik_outdated_info` only exists on `GET /servers/{uuid}` (shape verified live against v4.3.7), so each listed server is fetched individually; a failed detail fetch degrades to an `errors` entry rather than killing the scan. The `unhealthy_*`/`unreachable_servers` summary counts keep their pre-existing critical-only meaning; warnings are counted separately in `summary.warnings`.

## Live verification (Coolify v4.3.7)

New `issue-336.integration.test.ts` passes 6/6 against our production instance, and a direct `findInfrastructureIssues` run surfaces a real pending Traefik `3.6.8 → 3.6.23` patch plus 7 `running:unknown` resources, with criticals cleanly separated:

```text
{
 "total_issues": 15,
 "unhealthy_applications": 2,
 "unhealthy_databases": 0,
 "unhealthy_services": 5,
 "unreachable_servers": 0,
 "warnings": 8
}
warning  server  localhost — Proxy (Traefik) patch update available: 3.6.8 → 3.6.23
warning  application  … — Application running but health unknown (running:unknown) …
critical application  … — Application status: exited:unhealthy
```

Our estate doesn't reproduce the id collision (7/7 databases intact), so item 1's collision case is pinned by unit tests; the integration suite asserts the live invariant instead (`listDatabases` ⊇ standalone rows of `/resources`, no duplicates, every row typed).

## Validation

- `npm test` — 671/671 (was 655; 16 new)
- `npm run lint` — 0 errors (52 pre-existing warnings)
- `npm run format` / `prettier --check` — clean
- `npm run check:chunk-drift` — 14/14
- `npm --prefix evals run snapshots` — 4/4 (one intentional update: `environments` description)
- `npm run test:integration` (new suite) — 6/6 live

### DIFF
--- a/src/lib/coolify-client.ts
+++ b/src/lib/coolify-client.ts
@@ -426,6 +426,15 @@ function toDatabaseSummary(db: Database): DatabaseSummary {
   };
 }
 
+/**
+ * "running:unknown" — the container is up but Coolify has no health signal for
+ * it, usually a missing or broken healthcheck. Not an outage, but find_issues
+ * previously ignored it entirely (#336).
+ */
+function isRunningWithUnknownHealth(status: string): boolean {
+  return status.startsWith('running') && status.endsWith(':unknown');
+}
+
 function toServiceSummary(svc: Service): ServiceSummary {
   return {
     uuid: svc.uuid,
@@ -1394,12 +1403,48 @@ export class CoolifyClient {
   // Database endpoints
   // ===========================================================================
 
+  /**
+   * List databases, augmented from `/resources`.
+   *
+   * Coolify keeps a per-type id sequence and `GET /databases` merges the
+   * per-type collections keyed on that id, so two database types sharing ids
+   * silently shadow each other (verified live: 2 Postgres + 2 Dragonfly, both
+   * pairs ids 1 and 2, returns only the Dragonflys). `/resources` reports every
+   * database correctly, so rows it knows about that `/databases` dropped are
+   * merged back in by uuid. When `/databases` is complete the merge is a no-op;
+   * if `/resources` fails the plain `/databases` result is returned unchanged.
+   * @see https://github.com/StuMason/coolify-mcp/issues/336
+   */
   async listDatabases(options?: ListOptions): Promise<Database[] | DatabaseSummary[]> {
     const query = this.buildQueryString({
       page: options?.page,
       per_page: options?.per_page,
     });
-    const dbs = await this.request<Database[]>(`/databases${query}`);
+    const [dbsResult, resourcesResult] = await Promise.allSettled([
+      this.request<Database[]>(`/databases${query}`),
+      this.request<ResourceListItemFull[]>('/resources'),
+    ]);
+    if (dbsResult.status === 'rejected') throw dbsResult.reason;
+    let dbs = dbsResult.value;
+    if (
+      Array.isArray(dbs) &&
+      resourcesResult.status === 'fulfilled' &&
+      Array.isArray(resourcesResult.value)
+    ) {
+      const seen = new Set(dbs.map((db) => db.uuid));
+      // Every standalone-* resource type is a database; applications and
+      // services carry their own type strings.
+      const dropped = resourcesResult.value.filter(
+        (r) =>
+          typeof r.type === 'string' &&
+          r.type.startsWith('standalone-') &&
+          r.uuid &&
+          !seen.has(r.uuid),
+      );
+      if (dropped.length > 0) {
+        dbs = [...dbs, ...(dropped as unknown as Database[])];
+      }
+    }
     return options?.summary && Array.isArray(dbs) ? dbs.map(toDatabaseSummary) : dbs;
   }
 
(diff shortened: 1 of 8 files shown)


StuMason (dismissed):
This is exactly what #336 needed, and it is careful where it matters: twins surfaced but never deduped, the `/resources` merge degrades to the plain `/databases` result, merged rows go through the sanitizer boundary, and `list_databases` stays summary-projected. Pulled the branch: 672/672, lint clean, and the live verification against v4.3.7 catching a real pending Traefik patch is the kind of receipt most PRs never bring.

One non-blocking nit, inline: the merge assumes `/databases` returned the complete set, so a paginated call re-appends every database from the other pages. A one-line guard (skip the merge when `page` or `per_page` is set) covers it.

Approving as-is. Your call: happy to merge now and take the guard as a follow-up, or wait if you would rather push it onto this branch first. Say the word either way.

src/lib/coolify-client.ts
```diff
@@ -1394,12 +1403,48 @@ export class CoolifyClient {
...
+   * if `/resources` fails the plain `/databases` result is returned unchanged.
+   * @see https://github.com/StuMason/coolify-mcp/issues/336
+   */
   async listDatabases(options?: ListOptions): Promise<Database[] | DatabaseSummary[]> {
     const query = this.buildQueryString({
       page: options?.page,
       per_page: options?.per_page,
     });
-    const dbs = await this.request<Database[]>(`/databases${query}`);
+    const [dbsResult, resourcesResult] = await Promise.allSettled([
```
StuMason, line 1423: Non-blocking: this merge assumes `/databases` returned the complete set, so a call with `page`/`per_page` set will re-append every database from the other pages as "dropped" rows. In practice nobody paginates a database list, but a one-line guard keeps the contract honest:

```ts
if (options?.page === undefined && options?.per_page === undefined && dropped.length > 0) {
```

(or skip the `/resources` fetch entirely when paginating). Fine as a follow-up.

kashik0i: Verified: paginated /databases is being compared with the complete /resources set, so off-page rows can be appended. I’ll add the pagination guard and regression test on this branch, then push the update.

kashik0i: Fixed in 6e32e5e. The collision merge now runs only when neither page nor per_page is set, and the regression test keeps an off-page /resources row out of the paginated result. Local verification: 672/672 tests passed, lint completed with 0 errors, TypeScript build passed, and formatting passed.

kashik0i: Final head 38916e5 is updated with current main. Fresh verification on that head: 672/672 local tests passed and all 6 issue #336 live integration tests passed against Coolify 4.3.9. Ready for re-review.

StuMason: Guard and test are exactly right — resolved. Thanks for the quick turnaround.

StuMason (approved):
Re-approving the updated head: pagination guard + off-page test verified locally (672/672) and CI green.

PR #354, opened 2026-08-25
feat: expose destination_uuid on service create, database update, connect_to_docker_network, list_destinations (#351)

### LINKED ISSUE #351: Three Coolify API parameters not exposed: destination_uuid on service create, public port on existing databases, connect_to_docker_network
First, thanks for this server. I spent a full day driving a Coolify instance through it and the tool design held up well, `service`, `env_vars`, `deploy` and `scheduled_tasks run_once` in particular. I hit three gaps where the Coolify API supports a parameter that the MCP does not expose, and each one forced me out of the MCP and into the browser or the CLI.

Version tested: **2.19.3**, Coolify **4.3.9** self-hosted.

### 1. `service create` cannot set `destination_uuid`

On a server with more than one Docker network, `POST /services` requires `destination_uuid`. Without it the API answers:

```
Server has multiple destinations and you do not set destination_uuid.
```

The `service` tool has no such parameter, so creating a service is impossible on those servers. Note that `application` already exposes `destination_uuid`, so this looks like an oversight rather than a design choice. `destination list --server <uuid>` would also be useful to discover the value.

### 2. No way to expose an existing database on a public port

`database` only offers `create` and `delete`. `is_public` and `public_port` can be set at creation but never afterwards, while the UI lets you toggle it on an existing database at any time. That is the normal path for a one-off dump or an external client, and it currently has no MCP equivalent.

### 3. `connect_to_docker_network` is not exposed

The service-level toggle that attaches a stack to the shared `coolify` network has no parameter on `service update`. Without it, two services cannot resolve each other by container name, which blocks anything that needs cross-stack traffic, a database migration between two stacks for instance.

### Not a bug, for the record

I initially suspected the `logs` tool of sending the wrong query parameter, because the input is named `container`. Reading `coolify-client.js` shows it maps correctly to `sub_service_name`, and the official CLI fails identically on the same services. That one is a Coolify server-side bug, reported at coollabsio/coolify#11454. Your tool was faithfully surfacing their 500.

### Suggestion

The first gap is the blocking one. The other two are quality-of-life. Happy to test a build against a real instance if that helps.

StuMason:
Closes #351.

Three Coolify API parameters the tools never surfaced, reported by a field-tester on 2.19.3 against Coolify 4.3.9. Each claim was checked against the vendored OpenAPI spec before building; all three hold.

## What changed

| Gap | Fix |
|---|---|
| `service create` cannot set `destination_uuid` | `service` tool now takes `destination_uuid` (and `environment_uuid`) and forwards them to `POST /services`. The type already had the field; only the schema and handler were missing it. |
| No way to expose an existing database on a public port | `database` gains an `update` action over the existing `client.updateDatabase()` (`PATCH /databases/{uuid}`). Takes `is_public`, `public_port`, `public_port_timeout`, `limits_*` and credentials; create-only fields are stripped with a typed rest destructure. `is_public: true` goes through `guardDestructive`, so in HTTP mode a human confirms before a database becomes internet-reachable. |
| `connect_to_docker_network` not exposed | Added to `UpdateServiceRequest` and the `service` schema, alongside `is_container_label_escape_enabled` (previously documented as UI-only, wrongly). `service update` now sends an explicit typed pick instead of spreading every tool arg into the PATCH. |
| No way to discover a destination uuid | New read-only `list_destinations` tool: `GET /destinations`, or `GET /servers/{uuid}/destinations` with `server_uuid`. Pre-4.2 instances 404; that is mapped to a clear version hint. |

Tool count 44 → 45. Eval contract snapshots updated, fixture serves the destination routes, selection eval covers the discovery path. README, SKILL.md, CLAUDE.md and CHANGELOG updated.

## Backwards compatibility

Checked against upstream `ServicesController`/`DatabasesController` at each tag, since this project's job is the versions that predate Coolify's native MCP:

| Param | v4.0.0-beta.380 | v4.1.2 | v4.2.0 | v4.3.9 |
|---|---|---|---|---|
| `destination_uuid` on `POST /services` | ✅ | ✅ | ✅ | ✅ |
| `is_public`/`public_port`/`limits_*` on `PATCH /databases/{uuid}` | ✅ | ✅ | ✅ | ✅ |
| `connect_to_docker_network` on `PATCH /services/{uuid}` | ❌ 422 naming the key | ✅ | ✅ | ✅ |
| `GET /destinations` | ❌ 404 (mapped) | ❌ 404 (mapped) | ✅ | ✅ |

Nothing removed, every new parameter optional, both version-gated paths fail loud.

## Verification

- `npm test`: 732/732, lint 0 errors, `tsc` clean (src and evals), contract snapshots green.
- Live: `listDestinations()` against a **4.1.2** instance 404s on both routes, as expected from the table. I could not exercise the three mutating paths live (no multi-destination server here; not toggling public ports on production databases), so the reporter's offer to test a build is worth taking up.

## Branch note

Targets `v3` since 2.19.4 was the last 2.x feature release. The diff does not touch anything v3-specific, so it cherry-picks cleanly onto `main` if a 2.19.5 is wanted for stdio users on npm before 3.0 lands.


StuMason:
Addressed in the follow-up commit:

1. **`instant_deploy` regression** — back in the pick, plus `is_container_label_escape_enabled` (the stale "CANNOT be changed via API" note on `UpdateServiceRequest` is rewritten). Test asserts both forward.
2. **Limits** — `limits_*` and `public_port_timeout` added to the `database` schema so the description is honest. Credential/limits passthrough test added.
3. **Typed pick** — rest destructure of the create-only keys; `updateData` is now the typed remainder rather than `any`.
4. **`Destination`** — only `uuid`/`name`/`network` required.
5. **`is_public: true` guarded** — routed through `guardDestructive` with the database name and port in the prompt. Test: with `requireElicitation` and no elicitation capability it refuses and never calls `updateDatabase`; `is_public: false` is not guarded.
6. Smaller notes: 404 on `/destinations` maps to a "Coolify v4.2+" hint with the UI fallback; `environment_uuid` described (no local "one of" guard, to stay consistent with `application` create which doesn't have one either); fixture serves both destination routes; selection case added; CHANGELOG count fixed.

## Backwards compatibility (checked against upstream controllers at each tag)

| Param | v4.0.0-beta.380 | v4.1.2 | v4.2.0 | v4.3.9 |
|---|---|---|---|---|
| `destination_uuid` on `POST /services` | ✅ | ✅ | ✅ | ✅ |
| `is_public`/`public_port`/`limits_*` on `PATCH /databases/{uuid}` | ✅ | ✅ | ✅ | ✅ |
| `connect_to_docker_network` on `PATCH /services/{uuid}` | ❌ (422, named in the error) | ✅ | ✅ | ✅ |
| `GET /destinations` | ❌ 404 | ❌ 404 | ✅ | ✅ |

Nothing removed, every new param optional, and the two version-gated paths fail loud (Coolify's `$allowedFields` 422 names the extra key; the 404 is now mapped). 732/732 tests, lint clean, snapshots regenerated.

StuMason:
Follow-ups in 03eae89a:

- **#1 fixed.** Both halves of the expose-database prompt go through `sanitizeForPrompt`. Regression test feeds a `[Approve](https://evil.example)` name and a `") — **SAFE, routine** — (ignore` uuid through a stubbed `elicitInput` and asserts none of it survives into the dialog.
- **#2 fixed.** `/destinations` 404 hint now names the wrong-server-uuid cause, same shape as the `/tags` hint. Test on `listDestinations('nope')`.
- **#3 taken.** `database create` with `is_public: true` is guarded the same way as update. A comment saying create was out of scope would only have documented the bypass.
- **#4 taken as a warning**, not a guard: every credential field's `.describe()` now says update rotates live with no confirmation and breaks apps holding the old value.
- `database update` with nothing to change returns `Error: nothing to update` instead of `PATCH {}`.
- Negative `destination_uuid` test added for `service create`.
- `database` contract snapshot refreshed for the new descriptions.

Not taken: the `Destination` type note reads the comment wrong (it says everything *past the identity fields* is optional, which is what the type does); the fixture `split('?')` idiom matches the existing `/projects` route. `health_check_*` / `*_conf` on update and service-level `urls` are follow-up issues.

737/737, tsc, lint 0 errors, chunk-drift OK, snapshots 4/4.

StuMason:
Second pass in the latest commit:

- **#1 fixed**: README table and `evals/src/fixture/server.ts` re-prettiered; `format:check` clean in both roots.
- **#2 fixed**: `is_container_label_escape_enabled` forwarded on `service create` (it is in the `POST /services` body), description no longer says update-only. Test added.
- **#3 taken**: rotating any `*_password` field on `database update` now goes through `guardDestructive` with a prompt naming the fields being rotated. Pass-through on stdio, fail-closed in HTTP mode. Tests for the guarded case and for a limits-only update staying unguarded.
- **#4 fixed**: `service update` with nothing to change returns `Error: nothing to update`.
- **#5 fixed**: `manifest.json` / `server.json` say 45; the CLAUDE.md checklist now lists them.
- **#6 fixed**: `application` and `database` `destination_uuid` point at `list_destinations`. Engine-matching and "limits_* need a restart" notes added to `database`.
- `Destination` comment reworded to say why the identity fields are typed required.

Not taken: test-describe naming, fixture version, live integration test (follow-up). 741/741, tsc, lint 0 errors, chunk-drift OK, snapshots 4/4.

StuMason:
Third pass in the latest commit:

- **#1 fixed**: exposure and rotation now share one `guardDestructive` call whose summary names both, so `is_public: true` + `postgres_password` in one call reads "Expose … Also rotate postgres_password …". Test covers the combined call and asserts the new password never appears in the dialog.
- **#2 fixed**: `package.json` says 45; CLAUDE.md checklist lists it with the other two.
- **#3 fixed**: engine-matching note moved onto the `database` tool description.
- **#4 fixed**: `is_public` description covers create and update.
- **#5 fixed**: fallback is `(assigned by Coolify)`.
- **#6 taken in part**: the `*_user` fields join the guarded set. `image`, `public_port` and `limits_*` stay unguarded on purpose, and the `DB_CREDENTIAL_FIELDS` comment now says why (they are the ordinary reasons to call `update`; a dialog on every one teaches people to click through).

One thing worth knowing: the per-field rotation warnings from the last pass had pushed `tools/list` to 8074 tokens against the 8000 gate in `evals`. The warning is now stated once in the tool description instead, and the gate is back under. 742/742, tsc, lint 0 errors, prettier clean, chunk-drift OK, snapshots 4/4.

kashik0i:
Live-verified this PR at `17ec6ce` against local Coolify `4.3.12` through a fresh MCP stdio process.

Passed:

- `service create` forwarded `destination_uuid` and created a disposable service.
- `database update` changed `is_public: false -> true` and set `public_port: 55432`.
- `service update` changed `connect_to_docker_network: false -> true`.
- Build passed; unit suite passed `742/742`.
- Disposable service was deleted, Docker state was clean, and the seeded database was restored to its exact prior state.

One contract gap surfaced during rollback: MCP cannot clear `public_port` back to `null`:

```text
Input validation error: Invalid arguments for tool database:
public_port: Invalid input: expected number, received null
```

Coolify accepts this (`DatabasesController`: `numeric|nullable`), but this branch exposes `public_port` as `z.number().optional()` and `public_port?: number`. A narrow follow-up would be `z.number().nullable().optional()` plus `number | null` in the request type and a regression test that forwards `null` on update.

No product source changes made during verification. All temporary resources and mutations were cleaned up.

kashik0i:
Follow-up patch for the nullable rollback gap is now in #357. It keeps `null` update-only, adds real MCP transport coverage, updates the tool contract snapshot/docs, and passed the live stdio rollback probe.

PR #355, opened 2026-08-26
fix(deployments): honor app list pagination

ubranch:
## Problem

`deployment` accepts `page` for `list_for_app`, but the handler drops it. Coolify therefore receives no `skip`/`take` query and every requested page returns the newest 10 deployments. This makes historical incident correlation silently incomplete.

## Fix

- add `per_page` for `list_for_app`
- map `page` and `per_page` to Coolify's documented `skip` and `take` parameters
- preserve the existing unpaginated request when neither option is supplied
- add client and MCP handler regressions

## Verification

- red: focused tests showed page 2 still called the unpaginated endpoint and handler dropped both values
- green: 556 focused tests passed
- `npx tsc --noEmit` passed
- ESLint: 0 errors; existing warnings only

### DIFF
--- a/evals/src/contract/__toolsnaps__/deployment.json
+++ b/evals/src/contract/__toolsnaps__/deployment.json
@@ -22,7 +22,14 @@
         "type": "number"
       },
       "page": {
-        "type": "number"
+        "type": "integer",
+        "exclusiveMinimum": 0,
+        "maximum": 9007199254740991
+      },
+      "per_page": {
+        "type": "integer",
+        "exclusiveMinimum": 0,
+        "maximum": 9007199254740991
       },
       "max_chars": {
         "type": "number"
--- a/src/__tests__/coolify-client.test.ts
+++ b/src/__tests__/coolify-client.test.ts
@@ -3546,6 +3546,17 @@ describe('CoolifyClient', () => {
       );
     });
 
+    it('should paginate application deployments with Coolify skip and take parameters', async () => {
+      mockFetch.mockResolvedValueOnce(mockResponse({ count: 25, deployments: [] }));
+
+      await client.listApplicationDeployments('app-uuid', { page: 2, perPage: 10 });
+
+      expect(mockFetch).toHaveBeenCalledWith(
+        'http://localhost:3000/api/v1/deployments/applications/app-uuid?skip=10&take=10',
+        expect.any(Object),
+      );
+    });
+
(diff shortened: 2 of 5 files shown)


StuMason (commented):
Nice catch, and thanks for the tight PR. Checked the fix against Coolify's OpenAPI spec: `GET /deployments/applications/{uuid}` takes `skip` (default 0) and `take` (default 10), so the mapping is correct and the 10 default when only `page` is supplied matches upstream. Pulled the branch locally: 674/674 tests, lint 0 errors, tsc clean, chunk-drift OK.

Two CI notes:

- **Tool contract snapshots (hard gate)** is red because the `deployment` schema changed (`page` tightened to a positive int, `per_page` added). That gate is intentional: any schema change needs the snapshot refreshed and committed alongside. One command fixes it:

  ```
  npm --prefix evals ci && npm --prefix evals run snapshots:update
  ```

  Commit the resulting change to `evals/src/contract/__toolsnaps__/deployment.json` (it adds `per_page` and the integer/exclusiveMinimum constraints on `page`) and the gate goes green.

- **claude-review** cannot get an OIDC token on fork PRs. Not yours, ignore it.

Happy to merge once that snapshot commit is up.

ubranch:
The tool contract snapshot gate should be green now — I pushed `2a1f4ae` regenerating `evals/src/contract/__toolsnaps__/deployment.json`. The pagination fix narrows `page` to a positive integer and adds `per_page`, so the snapshot needed to record the new schema. Verified locally with `npm ci && npm run snapshots` in `evals/`: 4 passed, 0 failed.

Separately, the `claude-review` check on this PR fails for a reason unrelated to the change:

```
error: Unable to get ACTIONS_ID_TOKEN_REQUEST_URL env variable
Could not fetch an OIDC token. Did you remember to add `id-token: write` to your workflow permissions?
```

`.github/workflows/claude-code-review.yml` is missing `id-token: write` in its `permissions` block. I didn't touch it here since workflow changes from a fork PR need extra scope and are out of scope for a pagination fix.

The new checks are sitting in `action_required` — they need a maintainer approval to run.

StuMason:
Merged and out in v3.0.0, thank you for the regression tests and for regenerating the contract snapshot rather than leaving it to me.

An app list that does not fit in one page is fleet-sized usage, and that is one of the reasons #367 exists: one config with named instances, an `instance` argument on every tool, and `instance: "all"` for estate-wide health. If dubbe.ai runs more than one Coolify I would like your read on that proposal before I build it.

PR #357, opened 2026-08-28
fix: allow clearing database public port

kashik0i:
## Summary

Follow-up to #354's live mutation verification. Coolify accepts `public_port: null` on database updates, but MCP validation rejected it, so an assigned public port could not be cleared.

## Changes

- Allow `public_port: null` through the database tool schema and update request type.
- Reject `null` on database create so nullable behavior stays update-only.
- Document the clearing behavior and refresh the database tool contract snapshot.
- Add a real in-memory MCP transport regression covering both update and create.

## Test plan

- [x] Tests added/updated
- [x] Manually tested locally

- Root build passed.
- Root tests: 743/743.
- Root lint: 0 errors; existing warnings only.
- Spec drift, chunk drift, and Prettier checks passed.
- Eval contract snapshots: 4/4; eval typecheck/lint/format passed.
- Fresh stdio MCP call against local Coolify 4.3.12 forwarded `public_port: null`; raw API state remained `is_public: false`, `public_port: null`.

## Checklist

- [x] Documentation updated if needed


kashik0i:
CI note: build and scan passed. `claude-review` failed before any review ran because the fork PR job had no OIDC environment (`ACTIONS_ID_TOKEN_REQUEST_URL`; action asks for `id-token: write`). It posted no comments or findings, so this is workflow authentication, not a code/test failure.

StuMason:
Out in v3.0.0. That is seven merged PRs from you, and the field-test follow-ups in #349 are the most useful outside contribution this repo has had: the exact-match resolution rules and `find_issues` signals are now the baseline every other tool is held to.

3.1 is #367 (multi-instance) and #368 (`doctor`), both shaped by your reports. Two things I would like to know: what i-Mitigate is building on it, and whether the team ever needs hands on Coolify or MCP work beyond what lands here. I do contract days, and I would rather say that plainly than have you find out by accident.

PR #363, opened 2026-09-07
fix: clear npm audit high (fast-uri) blocking CI

StuMason:
The security-audit step is failing every PR (see #362's build 22.x) on 4 advisories that landed since 2.19.4 shipped: **fast-uri high** (SSRF / host confusion), qs + humanfs moderates, body-parser low.

This is `npm audit fix` plus review follow-ups — lockfile and CHANGELOG only, no code:

- **Scope, precisely:** this unblocks the `npm audit --audit-level=high` CI gate. It does not change what consumers install — downstream resolves `fast-uri@3` via `@modelcontextprotocol/sdk` → `ajv`, and our `overrides` only apply at this repo's root.
- **body-parser low left in place deliberately:** this server only constructs `StdioServerTransport`; the SDK's express paths never load.
- **Cooldown check done:** `@humanfs/types@0.15.0` and `es-define-property@1.0.1` are 2024 publishes, `qs@6.16.0` is 9 days old, `fast-uri@4.1.4` (5 days) is the advisory fix itself.
- **libc metadata restored:** the npm-10 rewrite dropped 16 `libc` fields; restored verbatim from main (no versions changed), `npm ci` under npm 11 validates clean.

Verified locally: audit gate exits 0, build + full test suite (674) green.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Security
+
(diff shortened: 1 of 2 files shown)


StuMason:
Addressed the review points:

- **Cooldown / provenance (1):** publish dates checked — `@humanfs/types@0.15.0` (Sep 2024) and `es-define-property@1.0.1` (Dec 2024) are old packages new only to this tree; `qs@6.16.0` is 9 days old; `fast-uri@4.1.4` (5 days) is the advisory fix itself, so its age is inherent. No attestations on any of them, but nothing inside the cooldown window beyond the fix we came for.
- **libc metadata (2):** restored all 16 fields. npm 11 would not re-add them on install because it reuses resolved lockfile entries without re-fetching manifests, and none of the affected packages changed version — so they were copied back verbatim from main and validated with `npm ci` under npm 11.
- **CHANGELOG (3) and scope wording (4):** added a `### Security` entry under `[Unreleased]` that says this unblocks the CI gate and explicitly does not change consumer installs (downstream gets `fast-uri@3` via the SDK → ajv), and notes why the body-parser low is left alone (stdio-only, express paths never load). PR body updated to match.
- **evals/site lockfiles:** deliberate omission, as suspected — neither workflow audits, and both get swept in the v3 packaging pass anyway.

Issue #291, opened 2026-07-23
Include whether an env variable is "preview" type or not


petteremil:
### Problem or motivation

It seems that the mcp doesn't let the llm know whether a variable is a preview type or not, which confuses it a lot, since it will think that variables are set wrong, then it corrects, often the wrong one, and gets really confused 😅

### Proposed solution

Add information about what type of env variable a variable is.
I'm not sure on the setting of variables, but I assume you'd need to be able to distinguish what you are setting as well :)

### Alternatives considered

Using the api directly works fine

StuMason:
Thanks for reporting this, and for describing the symptom rather than just the missing field. The "it thinks variables are set wrong, then corrects the wrong one" detail is what makes this worth fixing quickly, because a confidently wrong correction to an env var is a genuinely bad failure mode.

You are right on both counts, and I have confirmed where it goes wrong.

**Reading.** The `list` action returns a trimmed projection rather than the raw Coolify payload, to keep token usage down. That projection is `toEnvVarSummary` in `coolify-client.ts`, and it currently emits only `uuid`, `key`, `value`, `is_buildtime` and `is_runtime`. Coolify does return `is_preview`, and the `EnvironmentVariable` type already declares it, but the summary drops it on the floor before the model ever sees it. So the model is not ignoring the distinction, it is genuinely blind to it. That is the bug.

**Writing.** Your instinct here was right too. `is_preview` is currently only accepted on the `bulk_update` path's `data[]` array. The single `create` and `update` actions do not expose it, even though `CreateEnvVarRequest` and `UpdateEnvVarRequest` both already support the field. So today you cannot set a preview variable through the normal path, which is presumably why going at the API directly was easier.

## Fix

1. Add `is_preview` to `EnvVarSummary` and to the `toEnvVarSummary` projection, so `list` surfaces it. One boolean per variable is a negligible token cost against the confusion it prevents.
2. Add `is_preview` to the `env_vars` tool schema and thread it through the single `create` and `update` paths for applications, services and databases, matching the existing bulk behaviour.
3. Update the tool description so the model is told preview and production variables are distinct scopes and that a key can legitimately exist in both with different values. Without that, surfacing the flag alone might still leave it "helpfully" reconciling the two.

Point 3 is really the one that fixes your actual symptom, so thanks for describing the behaviour you saw and not just the missing field. Will link the PR here.

PR #364, opened 2026-09-07
fix: return exact environment values

Tchorizo:
## Summary

Fix `env_vars` exact-key + `reveal=true` so supported API responses return the requested value instead of an application summary. When Coolify withholds both value fields, the tool now reports a sensitive-read capability error.

Related: https://github.com/Tchori-Labs/infra/issues/137

## Changes

- Request the full application env-var representation for exact-key reveal, then filter to that key.
- Reject `reveal=true` without an exact key before calling Coolify.
- Report a clear capability error when the API/token omits `value` and `real_value`.
- Document the collection-only API limitation and `read:sensitive` requirement.
- Add regression coverage for returned values, filtering, unavailable values, and the no-key guard. Fixtures use non-secret placeholders.

## Test plan

- [x] `npm test -- --runInBand --coverage=false src/__tests__/mcp-server.test.ts -t "env_vars tool handler"`
- [x] `npx eslint --quiet src/lib/mcp-server.ts src/__tests__/mcp-server.test.ts`
- [x] `npx prettier --check src/lib/mcp-server.ts src/__tests__/mcp-server.test.ts README.md CHANGELOG.md`
- [ ] Full CI suite

## Checklist

- [x] Documentation updated

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -5,6 +5,12 @@ All notable changes to this project will be documented in this file.
 The format is based on [Keep a Changelog](https://keepachangelog.com/en/1.0.0/),
 and this project adheres to [Semantic Versioning](https://semver.org/spec/v2.0.0.html).
 
+## [Unreleased]
+
+### Fixed
+
+- **`env_vars` exact-key reveal now returns a value or a capability error** (Tchori-Labs/infra#137). Application lists request the full env-var representation when `key` and `reveal: true` are supplied, then filter to that exact key. Reveal without a key is rejected before the API call, and APIs/tokens that omit both `value` and `real_value` return a clear sensitive-read capability error instead of claiming success. Coolify exposes only the collection env endpoint; there is no supported per-variable GET or reveal query flag.
+
 ## [2.19.4] - 2026-08-22
 
 The field-test follow-ups from #336, contributed by the field-tester. The headline is an upstream bug: `GET /databases` can drop whole database types, and `list_databases` now repairs that from `/resources`.
--- a/README.md
+++ b/README.md
@@ -119,7 +119,7 @@ Set `COOLIFY_MCP_ELICITATION=off` to turn the confirmations off entirely. It exi
 
 Secrets are masked at the API boundary. A client granted "list" access never sees plaintext credentials unless you explicitly opt in with `reveal: true`:
 
-- **`env_vars`**: variable values return as `***`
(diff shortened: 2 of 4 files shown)


StuMason (commented):
Thanks for this — it's a real bug, well diagnosed, and the fix is the right shape. The summary projection keeps `value` but drops `real_value`, so an exact-key reveal could come back with no usable value even when the API had returned one. Requesting the full representation for that one path costs nothing (same endpoint either way — `summary` is client-side projection only), and failing loudly when Coolify withholds both fields is much better than claiming success. The pre-flight rejection of `reveal` without a key also matches this codebase's philosophy — the comment right above `filterByKey` already calls dumping every secret the failure mode.

I ran the branch locally: 677 tests green, no new lint issues. The red CI is our own `npm audit` gate, broken on main and fixed by #363 — nothing to do with your change; an update-branch after that merges will clear it.

Two asks before merge, both small:

**1. `.some()` → a guarded `.every()` in the capability check.** Coolify auto-creates a preview twin for every production application env var (`is_preview: true`), and `GET .../envs` returns both — so an exact-key filter legitimately yields two rows (see #291; we never dedupe them). With `.some()`, one row missing its value fails the whole read even when the other row carried the plaintext you asked for. Something like:

```ts
if (reveal === true && filtered.length > 0 &&
    filtered.every((v) => !hasReturnedValue(v.value) && !hasReturnedValue(v.real_value))) {
  throw new Error('…');
}
```

still catches the withheld case (v4.2 strips every row, not one of them). While you're there: `reveal + key` matching zero rows currently returns `[]` silently — a "key not found" error would save the caller a head-scratch on a typo, but that's optional. Related: worth knowing your downstream flow will usually get **two** rows back for one key (production + preview twin) — read `is_preview` rather than assuming a single row.

**2. Changelog placement.** The reveal-without-key rejection is a behaviour change, not a fix — it removes a previously supported (and tested — the #276 test you had to edit) capability. We're on board with the change, but it belongs under `### Changed` with a line saying bulk reveal is now rejected, so anyone who was doing an all-values read finds out from the changelog rather than from an error. The value-fix part stays under `### Fixed` as you have it.

Minor heads-up: #363 also adds an `[Unreleased]` section, so whichever lands second gets a trivial conflict there.

Nice test discipline, by the way — placeholder fixtures instead of realistic-looking secrets, and the no-key guard asserting the client was never called. Appreciated.

StuMason:
Main's audit gate is fixed (#363 merged), so once you've pushed the review tweaks an update-branch/rebase will give you green CI. Heads up that the `[Unreleased]` changelog section from #363 is on main now, so expect a small conflict there — your entry slots in alongside it.

StuMason:
Landed in #380 for 3.1 with your commit intact and credited (`740c6e5d`, authored by you) — the two review asks were folded on top: guarded `.every()` for the preview-twin case and the reveal-all rejection recorded under Changed. Thank you for a real bug well diagnosed; you'll see it in the 3.1.0 notes. Closing this in favour of #380.

Issue #368, opened 2026-09-08
3.1: doctor, a diagnostic that turns "it's broken" into a one-line fix


StuMason:
## Why

Most "coolify-mcp is broken" reports are the environment, not the server, and the server currently says nothing useful when that happens:

- A team in Brazil (pedrorezendefig/hospital-reunioes#312) got 401s, ripped the MCP out and rewrote 30 tool calls against the CLI. The cause was their macOS Keychain storing the literal string `${COOLIFY_ACCESS_TOKEN}`. One diagnostic line would have kept them.
- Clawith bundles this server (dataelement/Clawith#717) and loaded 0 of 42 tools because `npx` was not in the container. Silent.
- herdctl (edspencer/herdctl#182) passed the config in the wrong shape. Silent.
- @StreamlinedStartup audited the client line by line in #292 and predicted which calls Coolify v4.2 would break. That check should run itself.
- Tokens with the wrong abilities produce 403s deep inside a tool call instead of a plain "your token cannot deploy" up front. Coolify's own MCP now gates tools on token ability; we should read the same thing and say it.

## Proposal

1. `npx @masonator/coolify-mcp doctor` (CLI) and a `doctor` tool (MCP). Same checks, two front doors.
2. Checks, each with a pass, fail or warn and a one-line fix:
   - config found, which mode (single or instances), which env vars were read, and whether any value looks like an unexpanded `${VAR}`
   - Coolify reachable, HTTPS, response time
   - Coolify version and whether it is inside the supported range this release was tested against
   - token valid, token abilities, and which tools are therefore unavailable
   - endpoint drift: the read-only probe for the calls that v4.2 changed (#292) so we notice before a user does
   - runtime: Node version, `npx` present when launched by a host, HTTP mode port free
3. Output is human readable by default and JSON with `--json`. Never prints a token, a secret, or an env value; prints the variable name and "set" or "unset".
4. Every tool error that matches a doctor check ends with "run doctor" and the check name.

## Acceptance

- The four failure stories above each reproduce as a fixture and each produce the right line.
- Secret-masking evals cover doctor output.
- README quick start ends with "then run doctor".
- Release note credits the people who reported the failures.

StuMason:
Four refinements from digging into the failure stories against what the codebase can actually promise:

[...]

PR #374, opened 2026-09-08
Startup config self-check (#368 slice) + Cloudflare Access service tokens (#373)

### LINKED ISSUE #373: 3.0.x: support Cloudflare Access service tokens for the Coolify base URL
## Why

If your Coolify control plane sits behind Cloudflare Access (a completely reasonable way to protect a self-hosted dashboard), the MCP server cannot reach it: every API call gets a 302 to the Access SSO page, and in HTTP mode that failure surfaces two ways — every remote tool call dies, and **new OAuth authorizations die too**, because the tier-2 Coolify-token proof validates against `GET /teams/current` through the same blocked base URL. The server's `/healthz` stays green throughout, which makes it a miserable thing to debug.

This is not hypothetical: it took down the reference deployment (mcp.stumason.dev → coolify.stumason.dev) on 2026-09-08 when an Access policy changed. The wire signature is a 302 to `<team>.cloudflareaccess.com/cdn-cgi/access/login/...` with `service_token_status: false` in the redirect JWT.

Cloudflare's sanctioned machine-to-machine path is a **service token**: Access lets requests through when they carry `CF-Access-Client-Id` / `CF-Access-Client-Secret` headers and a policy accepts that token. The client has no way to send them.

## Proposal

1. Two env vars, exactly mirroring Cloudflare's names: `CF_ACCESS_CLIENT_ID` and `CF_ACCESS_CLIENT_SECRET`. When both are set, every request to the Coolify base URL carries the two headers. Both or neither; one alone is a startup config error (fits #368's startup self-check).
2. Headers are attached **only** to requests to the configured Coolify base URL — never to any other fetch (the future CIMD fetch in particular must not leak them; `fetchPublicJson` stays header-free).
3. The secret is a credential: masked everywhere the sanitizer masks, never logged (#370's no-arguments rule already covers it), never echoed by doctor (#368 prints "set"/"unset" only).
4. Docs: a "Coolify behind Cloudflare Access" section in docs/http-mode.md — create the service token, add the Access policy, set the two vars. Note the alternative that avoids the problem entirely when the MCP container runs on the same host: point `COOLIFY_BASE_URL` at the control plane over the internal Docker network so calls never traverse Cloudflare.

## Acceptance

- With the vars set, requests to the base URL carry both headers; a fetch to any non-base-URL host provably does not (test asserts header absence).
- One-var-set fails fast at startup with a one-line message.
- Masking eval: the secret never appears in any tool response or log line.
- docs/http-mode.md section shipped.

StuMason:
Two small 3.0.x slices, interlocked because the CF Access pairing rule lives inside the self-check.

## Startup self-check (first slice of #368)

Both entry points run `checkStartupConfig` before serving anything and print every problem to stderr in one pass:

- **Unexpanded `${VAR}` / `$VAR` placeholders** in any checked variable — the macOS Keychain failure that cost the hospital-reunioes team their whole integration (#368, story 1). Fatal.
- **Control characters in the token** (a pasted newline makes `fetch` throw before the request even leaves). Fatal. Surrounding whitespace warns.
- **Base URL shape**: unparseable or non-http(s) is fatal; a URL already ending in `/api`/`/api/v1` warns, because the client appends `/api/v1` itself and every call would 404.
- **CF Access pair**: one of `CF_ACCESS_CLIENT_ID`/`CF_ACCESS_CLIENT_SECRET` without the other is fatal, naming the missing one.

The iron rule from #368 holds and is tested: messages name the variable and the shape of the problem, **never the value**. HTTP mode folds these into its existing collect-then-refuse `problems[]` pattern; stdio prints warnings and exits on errors before the transport starts.

This is deliberately the smallest shippable slice of doctor — the CLI/tool front doors, reachability, version-range, abilities and drift probes stay on #368.

## Cloudflare Access service tokens (closes #373)

`CF_ACCESS_CLIENT_ID` / `CF_ACCESS_CLIENT_SECRET` (Cloudflare's own names) become the `CF-Access-Client-Id`/`CF-Access-Client-Secret` headers on every request to `COOLIFY_BASE_URL` and nothing else:

- They ride as `customHeaders` on `CoolifyClient`, which by construction only ever talks to the base URL.
- They're passed to `validateCoolifyToken`, the tier-2 proof-of-access fetch — the path whose failure killed new authorizations in the 2026-09-08 incident. Spread first, so extra headers can never displace the `Authorization` being proven (tested).
- A test asserts the docs-search fetch (the one off-estate fetch in the serving path) carries no headers even with both vars set.

Docs: new "Coolify behind Cloudflare Access" section in `docs/http-mode.md` (service token setup, the internal-Docker-network alternative, and the Livewire silent-save warning), config-reference rows, and a troubleshooting entry for the "every tool call fails while /healthz stays green" signature. Also fixes the stale instruction to deploy the retired `v3` branch — HTTP mode ships on `main` since 3.0.0.

## Tests

762 passing (24 new), coverage 94.5% statements / 100% on the new module. No tool schema changes, so contract snapshots are untouched.


StuMason:
All six findings addressed in c7f7a45, and #1 was a genuinely good catch — verified empirically against undici before changing anything, and it behaves exactly as the review predicted: trailing LF/space on the token is normalized away and works; leading space/tab survives as `Bearer  <token>`; only interior NUL/CR/LF throw.

1. **Whitespace model corrected**: interior NUL/CR/LF → fatal, leading whitespace on the token → fatal (it becomes part of the credential), trailing whitespace → silent (it works; flagging it would refuse startup for `--env-file`/k8s-secret configs that are fine today). Tests updated to assert the trailing case is clean.
2. **`hasControlChars` replaced** with a NUL/CR/LF-only check, applied to the token *and* both CF Access values (interior line break in either now fails fast with the variable named).
3. **Wiring tests added**: `createHttpApp` → `validateCoolifyToken` customHeaders pass-through is now proven via a full authorize POST asserting the mocked `/teams/current` fetch saw both CF headers and the presented token — deleting the call-site argument now fails the suite. The stdio merge moved into a tested helper (below), so that call site is one trivial line.
4. **`--header` override is case-insensitive**: merge extracted to `mergeCfAccessHeaders`, which drops any env-derived key whose lowercase form appears in the CLI headers — no comma-joined duplicates. Tested.
5. **README line added** for the CF vars (works in both modes), pointing at the http-mode.md section.
6. Nits: `${VAR}` message now says what to do if a value legitimately contains `${`; URL message names the missing-scheme cause; `/api/v1` suffix promoted to fatal (guaranteed 404) with bare `/api` staying a warning; http.ts prints warnings only after the fatal gate; `docs/V3.md` opener retired.

769 tests passing.

StuMason:
Round two addressed in 671622b:

1. **Fixed — and fair cop, this one was mine**: moving the warning loop below the exit block made warnings unreachable on any fatal error, recreating the one-problem-per-boot loop. Warnings now print before the fatal gate in http mode, matching index.ts.
2. **Partially deferred**: the stdio config-building already moved into tested functions this PR (`mergeCfAccessHeaders`, `checkStartupConfig`), so both entry points are now thin call sequences of covered code. A proper entry-point harness (spawn or process.exit spies) is coming with the doctor PR, which restructures index.ts argv handling anyway — I'd rather build it once there than twice.
3. **Fixed**: `checkStartupConfig(env, transport)` — stdio no longer checks `MCP_PUBLIC_URL` at all, so a shared .env with an unexpanded Coolify magic var can't stop a stdio server that would run fine. Tested both directions.
4. **Fixed**: `parseHeaders` now replaces a repeated header case-insensitively (keeping the last spelling), so two flag spellings can no longer reintroduce the comma-joined duplicate. Tested.

771 tests passing.

Issue #369, opened 2026-09-08
3.x: respect Coolify token abilities per tool, instead of trusting the token


StuMason:
## Why

Coolify's built-in MCP gates every tool on the token's actual abilities — `ResolvesTeam::ensureAbility()` in `app/Mcp/` checks `$token->can('root') || $token->can($ability)`, lifecycle tools require the `deploy` ability, and member-role users are hard-blocked from sensitive scopes regardless of token abilities. We do none of that: whatever the token can do, every tool will try, and the user finds out via a 403 deep inside a tool call (#368 tells the same story from the diagnostics side).

Gating up front is better for the model too: a token without `deploy` should produce a server that refuses those tools with one plain sentence, not an agent that retries a 403 three ways.

## The blocker to resolve first

Upstream can gate because it lives inside Coolify and holds the Sanctum token object. From outside, we need actual introspection: an API endpoint that returns the current token's abilities. As far as we know **that endpoint does not exist** — verify, and if it's missing, open an upstream feature request (current-token abilities on `/teams/current`, or a `/tokens/current`). Coolify's own MCP proves the ability data is right there server-side.

## Proposal

1. Verify or request upstream introspection; link the upstream issue here.
2. Until introspection exists, the honest inference ladder (same as #368's): `read:sensitive` is provable (fetch an env var, check whether a value came back); write/deploy abilities are not provable without side effects and report as unknown.
3. Once introspection lands: read abilities at startup (and in doctor), fail fast with "your token lacks the X ability" on gated tools, and get `MCP_READONLY`-style filtering for free on read-only tokens.
4. Never gate by guessing. A wrong "you can't do that" is worse than a 403.

## Acceptance

- Upstream introspection verified present or requested (issue linked).
- With a read-only token: deploy/write tools fail fast with the one-line ability message, before any API mutation attempt.
- Masking evals unchanged; doctor (#368) reports abilities from the same source.

StuMason:
Blocker resolved — verified against upstream `main` at `1e8a27b` (identical in v4.3.18):

[...]

Issue #367, opened 2026-09-08
3.1: multi-instance (fleet) support, one config, every tool takes an instance


StuMason:
## Why

coolify-mcp is one token, one instance. The people running it hardest are running fleets:

- @mediafill asked for multi-server in #164 and has asked Coolify itself for the same thing twice (coollabsio/coolify#6936, #9980).
- @ubranch's pagination fix in #355 only exists because an app list did not fit in one page.
- @daniel-rudaev runs "many resources across multiple projects" and wrote a whole skill documenting the workarounds.
- @follox42 spent a day driving an instance through it and hit the edges (#351).
- @kashik0i's field-test PRs (#349, #357) are what a real estate looks like.
- @rudo50647 fires dbt and ingest jobs through it as scheduled tasks (#208).

Agencies with a Coolify per client, and anyone with prod and staging on separate instances, currently run two copies of the server with two configs. Several newer servers have shipped "fleet" mode as their headline feature. This is the one thing they have that we do not, and it is the thing that makes an agency choose a tool.

## Proposal

1. Config accepts a list of instances, each with a name, URL and token: env `COOLIFY_INSTANCES` (JSON) alongside the existing single-instance vars, which keep working unchanged.
2. Every tool gains an optional `instance` argument. Omitted means the default instance (the single-instance config, or the first in the list). Wrong name returns the list of valid names.
3. `list_instances` tool: name, URL, reachable, Coolify version, token abilities. No secrets in the output, ever (see #159, #271, #328 for why that sentence is in bold everywhere in this repo).
4. `find_issues` and `list_resources` accept `instance: "all"` and return results keyed by instance, so "what is unhealthy across my estate" is one call.
5. Resolution by name works per instance; the exact-match rules from #349 apply unchanged.
6. HTTP mode (v3.0.0) serves the same instance list; OAuth scopes are unchanged.

## Acceptance

- Existing single-instance configs run with zero changes and zero new output.
- A two-instance config in the eval suite: every tool called with `instance` set, wrong name, and omitted.
- Secret-masking evals run against every instance-aware tool.
- README section: "Running a fleet", with the agency-per-client example.
- Release note names the people above.

If you are one of the people tagged and your setup does not fit this shape, say so here before it is built.

StuMason:
[...]

Issue #292, opened 2026-07-25
Heads up: Coolify v4.2 breaks several client calls (405s + hidden secrets)


StreamlinedStartup:
# Heads up: Coolify v4.2 breaks several client calls (405s + hidden secrets)

Coolify v4.2 was merged into `v4.x` on Jul 19, 2026 via [coollabsio/coolify#10872](https://github.com/coollabsio/coolify/pull/10872). It is not on the releases index page yet, so it is easy to miss.

I went through `src/lib/coolify-client.ts` at `e8dfb4b` (current `main`, dated Jul 16) and found four areas that will break once users upgrade. Opening this as a heads up rather than a bug report, since nothing is broken today.

---

## 1. State-changing endpoints are now POST-only (hard 405s)

This is the headline breaking change in v4.2. Start, stop, restart, deploy, enable, disable, and server validate all require `POST`. A `GET` request now returns `405 Method Not Allowed`.

Four call sites in `coolify-client.ts` still use GET:

| Line | Client method | Problem | MCP tool affected |
|------|---------------|---------|-------------------|
| 710 | `validateServer` | No `method` passed, so `request()` defaults to GET | `validate_server` |
| 1176 / 1182 / 1189 | `startService` / `stopService` / `restartService` | Explicit `method: 'GET'` | `control` with `resource: 'service'` |
| 1268 | `deployByTagOrUuid` | Explicit `method: 'GET'` | `deploy`, plus the deploy-and-wait path |
| 1855 / 1859 | `enableApi` / `disableApi` | Explicit `method: 'GET'` | `system` actions `enable_api` and `disable_api` |

These call sites are already correct and need no change:

- Application start, stop, restart (lines 938 to 956) already use POST
- Database start, stop, restart (lines 1050 to 1062) already use POST
- `validateCloudToken` (line 1428) already uses POST

One thing to double check on `restartService`: it currently passes `latest=true` as a query string. That should still work as a POST query param, but it is worth confirming v4.2 did not move it into the request body.

Full list of endpoints that changed, from the upstream PR:

```
/enable
/disable
/deploy
/servers/{uuid}/validate
/applications/{uuid}/start
/applications/{uuid}/restart
/applications/{uuid}/stop
/databases/{uuid}/start
/databases/{uuid}/restart
/databases/{uuid}/stop
/services/{uuid}/start
/services/{uuid}/restart
/services/{uuid}/stop
/services/{uuid}/applications/{app_uuid}/start
[...]

PR #375, opened 2026-09-08
doctor CLI (#368): environment diagnosis with one-line fixes

StuMason:
**Stacked on #374** (it builds on the startup self-check module) — merge that first; this retargets to main after. Do not delete the base branch while this is open.

## What

`npx @masonator/coolify-mcp doctor` (add `--json` for machine output), the CLI front door from #368. Seven checks, each pass/warn/fail/skipped/inconclusive with a one-line fix:

- **config** — required vars present + the startup self-check shapes (unexpanded `${VAR}`, pasted line breaks, doubled `/api/v1`)
- **reachability** — unauthenticated probe with response time, `redirect: 'manual'` so **Cloudflare Access interception is named by its 302 signature** with the service-token fix attached
- **token** — accepted / rejected / valid-but-lacks-read, discriminated by status and the upstream 403 body
- **version** — against the tested range (4.0.x – 4.3.x); outside is a warning, flagged as the first suspect when something else is red
- **abilities** — read proven by the token check; **write and deploy probed side-effect-free** via `GET /enable` and `GET /deploy` with no params (middleware answers before any controller runs). Upstream has no token-introspection endpoint (verified on #369), so the probe is keyed on the `Missing required permissions` 403 body — never on 405, because pre-4.2 instances answer these GETs differently — and Coolify's Member-role hard block is distinguished from a missing ability by its distinct body
- **api-shape** — does the routing catch-all still return the `docs`-keyed 404 our v4.2 method-fallback detection reads? @StreamlinedStartup's #292 audit, running itself
- **runtime** — node version

Fleet-shaped per #367: checks run per instance over a list that today has one entry.

## Secret safety

Reports contain variable names and statuses, never values — enforced by sentinel tests across the healthy, rejected-token and network-failure paths, plus a test that no probe ever sends an unexpanded placeholder as a credential.

## Field-verified against the live estate

Run from the sandbox against the real control plane (which sits behind Cloudflare Access), it produced exactly the diagnosis that took us a multi-hour incident to reach yesterday:

```
coolify-mcp doctor — instance: default
  ✓ config: COOLIFY_BASE_URL set, COOLIFY_ACCESS_TOKEN set
  ✗ reachability: Cloudflare Access is intercepting requests to Coolify (redirect to the Access login page in 104ms)
      fix: Set CF_ACCESS_CLIENT_ID and CF_ACCESS_CLIENT_SECRET (an Access service token), or point COOLIFY_BASE_URL at the internal address — see docs/http-mode.md
  - token: blocked by an earlier failure
  ...
```

## Still on #368 after this

The MCP `doctor` tool (schema change → contract snapshot + eval churn, wants its own PR), "every matching tool error ends with run doctor", and secret-masking evals over doctor output.

784 tests passing (13 new). README quick start now ends with "then run doctor"; changelog credits the four failure reports that shaped the checks.


StuMason:
All findings addressed in ca444fb. Finding 1 was a genuine correctness bug and the review's reasoning was exactly right — the `api-shape` check in the same run proves the mechanism that defeated the `/enable` probe. What changed:

1. **Write probe dropped, honestly.** No write-gated GET exists anywhere in the API, so no side-effect-free write probe *can* exist. Doctor now reports `write: not probeable without side effects`, keeps the sound `/deploy` probe (GET-routed in every era, 400s without params), and when deploy itself can't be determined it says `deploy: could not determine` instead of silently shrinking the granted list. The fixtures that encoded the wrong model are corrected — `/enable` now falls to the catch-all in the healthy mock, like a real ≥4.2 instance.
2. **`--header` honoured**: doctor builds instance headers via the same `mergeCfAccessHeaders(env, parseHeaders(argv))` the server uses, with a test asserting every probe carries the flag value. No more confident false failures for proxy-auth setups.
3. **`process.exitCode` + return** — piped `--json` output can no longer be truncated.
4. **Version echo bounded**: a 200 that doesn't match `/^v?\d+\.\d+\.\d+/` is `inconclusive` ("…is COOLIFY_BASE_URL pointing at Coolify?") with nothing echoed — tested with an SSO-page body carrying a sentinel CSRF token. Matching answers are sliced to 32 chars. Error messages are credential-masked (`//user:pass@` → `//***@`) and capped at 140 chars.

Smaller things, all taken: unknown probe answers named in the detail; 401/5xx count as unknown, never granted; version range compares both bounds as tuples; `tokenOk` is its own boolean (a 200-empty-body can no longer produce `token: pass` + `skipped — token check did not pass`); the always-true guard is gone; `runtime` moved to a top-level `checks` array on the report so the JSON shape survives #367's fleet; `doctor --help` prints usage (probe timeout documented there and in the README); the 10s-per-probe budget is in the README. `--timeout` left out for now — flag it if you disagree.

795 tests passing, doctor.ts at 100% lines. Re-verified live: same one-line Cloudflare Access diagnosis, `--help` exits 0.

StuMason:
Round two, triaged per the stopping rule (fix bugs, decline polish):

**Fixed (real bugs) in 3db28f3:**
1. `ok` now requires no failures AND no inconclusives, and the summary reports counts (`N failure(s), N warning(s), N inconclusive`) instead of a binary — the SSO-page case now exits 1 as it should. Warnings still don't gate.
2. The deploy probe treats the catch-all 404 (docs-key body) and any redirect as `unknown`, never `granted` — same-mechanism regression guard as the `/enable` fix, with a test that a controller's genuine 404 still counts as gate-passed.
3. `deploy: could not determine` is now status `inconclusive`, which with (1) stops it counting as green.

**Verified, no change (4):** measured rather than assumed — full 4-probe run against a local stub exits in **0.3s** (and 0.35s on the live CF-blocked path). undici's global agent doesn't hold the event loop open, so no body-cancel plumbing needed. If a field report ever says "doctor hangs", that's the first place to look.

**Deferred (5):** the substring coupling between doctor's gates and startup-check's message prose is a fair structural point but produces no wrong behaviour today. It lands naturally with the MCP doctor-tool follow-up on #368, which will restructure `StartupCheckResult` anyway — filed there rather than churning this PR. Same for the `configBroken` over-blocking edge (malformed `MCP_PUBLIC_URL` under `MCP_TRANSPORT=http` skipping Coolify probes): real but tiny, same follow-up.

797 tests passing. This PR is done from my side — further rounds get replies, not commits, unless something is an actual defect.

StuMason:
Both taken in b17a6e5 — they cleared the "actual defect" bar. A 429 from throttle middleware reading as `deploy: granted` was a today-reachable false grant on any shared instance, and the drifted catch-all copy was the same class. `probeAbility` is now an allowlist (2xx / paramless 400 / v4.2 405 / controller 404 prove a handler; everything else is unknown), and `isRoutingCatchAllBody` lives once in `api-shape.ts`, used by the client's method fallback, the probe, and the api-shape check — which now genuinely verifies what the client does. Tests cover the 429, the bare "Not found." second signal, and the docs-key case.

797 passing. That's the line for this PR: the evidence model is now allowlist-everywhere and shared-predicate — further hardening ideas go to the #368 follow-up.

StuMason:
Round four triage, in a742a49:

**Finding 1's premise is wrong, verified at source — but half of it was still worth taking.** `routes/api.php` at both **v4.2.0 and current main** registers `GET /deploy` → `OtherController::post_required` **behind `api.ability:deploy`**:

```
Route::get('/deploy', [OtherController::class, 'post_required'])->middleware(['api.ability:deploy']);
Route::post('/deploy', [DeployController::class, 'deploy'])->middleware(['api.ability:deploy']);
```

So on v4.2+ the GET probe is not a method rejection: the ability gate runs, a missing ability 403s with the parseable body, and a granted one reaches the 405-returning handler. Upstream kept GET registered precisely so the method change is discoverable. 405 stays in the granted set, now with that receipt in the comment. (The healthy v4.2+ instance therefore neither 404s nor false-fails — the two-branch dilemma doesn't arise.)

**Taken anyway, because the caution generalizes:** an undeterminable deploy (proxy rewrite, 429, future route move) now **warns instead of gating `ok`** — your `write` precedent, so doctor can never exit 1 on a healthy instance over an unprobeable extra ability. The gating `inconclusive` stays reserved for checks that verify the setup itself.

**Finding 2 taken**: `response.ok` is out of the granted allowlist — a paramless `GET /deploy` never legitimately 200s, so a 2xx reads as an anomaly, not proof. Tested.

**Finding 3 taken**: the GETs-only invariant is now a test over every probe call.

798 tests passing. That's the merge-ready line.

PR #378, opened 2026-09-09
Fleet mode: one server, several Coolify instances (#367)

StuMason:
Part of #367 — the core fleet surface. `instance: "all"` fan-out on `find_issues`/`list_resources`, the instances file registry and `tokenCommand` credential sourcing are follow-up slices.

## What

`COOLIFY_INSTANCES='[{"name":"staging","url":"https://…","token":"…"}]'` next to the existing vars (which define the default). With more than one instance:

- **Every tool takes an optional `instance`** — omitted = default; unknown name = rejected with the valid list *before* any request goes out.
- **`list_instances`** (fleet-only tool): name, URL, default flag, live Coolify version or the error reaching it. Never a token.
- **Every destructive confirmation names the instance**: `stop_all_apps on instance "staging"` / `Instance: staging` in the prompt body. Tested end-to-end through elicitation.
- **One `CoolifyClient` per instance** — the v4.2 method-fallback cache, version cache and sanitizer are client-scoped, so prod on 4.1 next to staging on 4.3 keep separate answers.

## The wall that shaped it

The single-instance `tools/list` sits at **~7.7k tokens against the 8k eval gate**; an `instance` property on 45 tools costs ~900. Rather than raise the gate (charging every single-instance user +11% per connect for a feature they don't use), **the property only exists when ≥2 instances are configured**. That makes "existing configs run with zero changes and zero new output" literally true — the 45 single-instance contract snapshots are byte-identical in this PR — and the fleet surface gets its own roster snapshot and its own 9k budget in `evals/src/contract/fleet.test.ts`.

## How routing works

`defineTool` is the one place all registrations pass through; it now resolves `args.instance` and runs the handler inside an `AsyncLocalStorage` context that a `client` getter reads. The 195 `this.client` call sites are untouched, and concurrent calls against different instances can't cross (tested with `Promise.all` across hosts). `guardDestructive`, the one confirmation chokepoint, scopes label and summary to the instance.

## Trust model (decided on #367)

A fleet is **one trust domain**. HTTP mode's tier-2 proof of access validates against the default instance only; every instance in a list is assumed to belong to the same owner. **Agencies with a Coolify per client run a server per client.** README and http-mode.md say this in bold. Per-instance OAuth scopes are 3.2.

## Not in this PR

- `instance: "all"` fan-out (find_issues / list_resources / find) — next slice.
- File-based registry + `tokenCommand` (the devrim/clezcoding prior art) — later slice.
- Doctor iterating the registry — 10-line follow-up once #375 lands (it's fleet-shaped already).

## Tests

786 unit (15 new: registry parsing incl. every malformed shape and a no-token-in-errors sweep; single-instance servers have no fleet surface; fleet schema on every tool; annotations table ↔ fleet server exactly in step; routing by name, default, unknown; concurrency isolation; list_instances output; confirmation naming; read-only fleet). Evals: 8/8 contract incl. the new fleet suite; single-instance snapshots unchanged.


StuMason:
One fix round, triaged (this PR's only one):

**Fixed — verified defects:**
1. **`instance` leaked into request bodies** — and wider than one handler: `application` update, `database`, `github_apps` and `database_backups` all rest-spread args into a body. Now stripped once at the `defineTool` chokepoint before any handler runs, with a regression test that drives `database_backups create` through fleet mode and asserts the POST body carries no `instance` key (and hits the right host). Future `...rest` handlers are covered by construction.
2. **Phantom localhost default** — stdio's localhost fallback now applies only when `COOLIFY_INSTANCES` is unset, so a lingering token can't synthesize a default that swallows every unqualified call.
3. **`list_instances` no longer takes `instance`** (5a) — a wrong name can never break the tool that corrects wrong names. Snapshot updated; unknown keys are dropped by the schema, tested.

**Deferred to the doctor-over-registry follow-up** (lands once #375 merges): per-entry shape checks for `COOLIFY_INSTANCES` (3) and a per-instance deadline on `list_instances` (4, pre-existing no-timeout across the client). Both real, neither wrong behaviour today.

**Declined:** case-insensitive names / `z.enum` (the error self-corrects and enum costs the budget the design exists to protect); the `RawInstance` typing nit; entry-point wiring tests (same answer as #375 — the logic is in tested functions). Trust model: on the record as deliberate, thanks for flagging it that way.

788 tests, 8/8 evals contract, single-instance snapshots still byte-identical.

PR #382, opened 2026-09-09
docs: 3.1 restructure — README front door, docs/ reference, site links out + llms.txt

StuMason:
Docs and site bump for 3.1: less prose, better structure. No server code changes.

## README is the front door only

Three install paths (one-click, local, remote), "then run doctor", a paragraph each for what it does, fleet and safety, every one linking out. The v2-era `## Design` token pitch is gone; the "45 tools" claim is now checked, not remembered.

## `docs/` is the reference

- `docs/tools.md`: the full table (moved from README), how the surface is shaped, version compatibility, the upstream gotchas the tools absorb.
- `docs/fleet.md`: `COOLIFY_INSTANCES`, what changes with two or more instances, the one-trust-domain rule (previously split between README and `http-mode.md`, each pointing at the other).
- `docs/doctor.md`: every check, what it proves, exit codes, what it refuses to guess.
- `docs/security.md`: the elicitation and masking write-ups, verbatim from the old README, plus how to report a vulnerability.
- `docs/http-mode.md`: kept its name (release notes and issues link to it); fleet section now defers to `fleet.md`.
- `docs/V3.md` → `docs/history/V3.md`.

## Tool count has one source of truth

`scripts/check-tool-count.mjs` (CI step, `--fix` rewrites) compares every "N tools" claim in README, CLAUDE.md, `package.json`, `server.json` and `manifest.json` against the evals roster snapshot: the tools a default install actually exposes, gated by the contract test. The site build reads the same roster; it was counting `defineTool` registrations, which since 3.1 includes the fleet-only `list_instances` and would have told single-instance visitors "46".

## Site

Single page, links out, as agreed.

- Install now covers remote/OAuth and fleet, and ends with doctor. The "Nothing to host. Your token never leaves your machine." line was false for two of three modes.
- New Docs section linking the reference; sixth capability tile for fleet (3+3 grid).
- JSON-LD `softwareVersion` read from `package.json` at build. It said `2.17.0`.
- `/llms.txt` and `/llms-full.txt` generated from README and `docs/*.md` at build (`?raw` imports), so they cannot drift from the repo. `Dockerfile.docs` and `.dockerignore` updated so the build context carries them; the site workflow now triggers on those paths and asserts both files and the version.
- `/docs` redirects to the GitHub docs tree.

Verified locally: site `check`/`test`/`build` green, `llms.txt` shows v3.1.0 and 45 tools, JSON-LD shows 3.1.0, `check:tool-count` in step, root prettier/lint clean, markdownlint clean.

| Desktop | Mobile |
| --- | --- |
| desktop | mobile |

## Also

`SECURITY.md` added, and **private vulnerability reporting is now enabled on the repo** (one settings toggle; the advisory link in `docs/security.md` needed it).


StuMason:
Review round applied in b536e27f (one round, per house rule):

1. **npm eats `--fix`**: added `check:tool-count:fix`; CLAUDE.md and CONTRIBUTING point at it.
2. **Fix-mode exit**: unmatched claim patterns are now "unfixable" and exit 1 in both modes; "fixed" only prints when everything was rewritten. Replacer rebuilds from match offsets instead of first-occurrence replace. Exercised locally: check exits 1 on a 44, fix rewrites it to 45 but still exits 1 while the table is broken.
3. **Table gated**: every roster tool must appear as backticked name in docs/tools.md; verified failing on a renamed `search_docs`, passing on the real table.
4. **site.yml** now triggers on `package.json`; header comment updated.
5. `llms.txt` uses `import.meta.env.SITE`; `llms-full.txt` globs `docs/*.md` keyed off `DOCS` and throws at build on a listed doc with no file (changelog deliberately excluded, said in the comment).
6. Nits: dead `rel=noopener` dropped; the Install lede now says where the token lives in each mode instead of the vaguer line.

Left as-is: sitemap listing the two txt routes (harmless), and site unit tests for the define-injected modules (the workflow greps cover the outputs; a vitest `getViteConfig` setup is a separate PR if wanted).

PR #388, opened 2026-09-09
feat: set the server instructions field (#339)

### LINKED ISSUE #339: Set the server instructions field — it is what tool-search clients read first
We do not set the server `instructions` field at all. That is now the single highest-leverage string in the server.

## Why it changed

Claude Code enables tool-definition deferral by default (`ENABLE_TOOL_SEARCH`, on for Sonnet/Haiku/Opus 4.5+). At session start it loads **only tool names and the server's `instructions` field**; full definitions are fetched on demand. Claude Code also states it imposes no fixed per-server tool cap.

So on that client the ~6,600-token `tools/list` achievement is largely not being paid at session start anyway — and what the model actually reads first is a field we leave empty. The token-budget discipline is still right for other clients, but `instructions` is now where the routing happens.

Anthropic's own research on progressive disclosure puts one worked example at 150k → 2k tokens, with the conclusion that upfront tool definitions are the dominant context cost and the fix is incremental loading plus good orientation text.

## What to put in it

Not marketing. Orientation the model cannot infer from tool names:

- The consolidation rule — every tool takes an `action`, call one with no arguments and it lists what it accepts
- Smart lookup — `diagnose_app` takes a UUID, name **or** domain; `diagnose_server` takes UUID, name or IP
- Where to start — `get_infrastructure_overview` before drilling
- That `list_*` returns projections and `get_*` returns detail
- That `_actions` carries the next call, so it should be followed rather than guessed at
- The destructive/confirmation boundary
- The version-sensitive bits worth knowing before a call fails

## Constraints

Directory review rules apply to this text as much as to tool descriptions: no instructing Claude to call other tools as a behavioural directive, no hidden or encoded instructions, no promotion. *"Describe what the tool does. Do not tell Claude how to behave."* Orientation, not personality.

## Tasks

- [ ] Draft `instructions`
- [ ] Add to the contract snapshot so it cannot drift unreviewed
- [ ] Add a selection eval case that depends on something only stated there
- [ ] Include its token cost in the budget assertion

Refs: https://www.anthropic.com/engineering/writing-tools-for-agents · https://www.anthropic.com/engineering/code-execution-with-mcp

StuMason:
Closes #339.

The `initialize` response now carries `instructions`. Claude Code loads tool names and this field at session start and fetches full definitions on demand, so until now the first thing the model read from this server was an empty string.

**What it says** (single-instance text is ~425 tokens; fleet adds one paragraph):

- tools are consolidated by `action`, and each `action` enum is the full list
- `list_*` returns projections, `get_*` returns one resource in full
- `diagnose_app` takes uuid, name or domain; `diagnose_server` uuid, name or IP
- `_actions` and `_pagination` are computed by the server from the real response
- read-only tools change nothing; destructive actions ask the person first, and a decline returns a message rather than an error; secrets are masked, `reveal: true` needs a `key`
- tested range 4.0.x – 4.3.x (from the same constant doctor uses), the 4.2-only calls, and what a 405 means
- fleet mode: `instance`, the default name, `list_instances`, instance-named confirmations
- read-only mode: says so instead of describing a boundary that does not exist

**What it does not say.** No behavioural direction, no promotion, no resource names, tokens or URLs. A unit test rejects "you must / you should / always call / never call" wording, and the text is identical for every user of the same mode.

**Wiring.** `buildInstructions()` in `src/lib/instructions.ts`, called from the constructor with the registry and options it already has. `instructions` rides `initialize`, not `tools/list`, so the single-instance tool list is byte-identical and the tool-list budgets are untouched.

**Contract.** Snapshotted in `evals/` as `_instructions.txt` and `_instructions.fleet.txt`, with its own budget assertion, so an edit shows up in review as a diff like a tool description would.

**Not done from the issue's task list:** a selection eval case that depends on something only stated here. The eval harness hands the model every full tool description, so no case can isolate the instructions field; it would measure the descriptions, not this. Noted rather than faked.


StuMason:

1. **Confirmation text is mode-accurate.** `requireElicitation` is now passed from the constructor. stdio: "ask the person for confirmation in clients that support elicitation … In a client without elicitation they run unconfirmed." HTTP: "require the person to confirm through elicitation … in a client without elicitation they are refused with a message rather than run." Read-only keeps its own sentence.
2. **Masking is its own paragraph in every mode**, names `get_application`, `get_database` and `get_service` as accepting `reveal: true`, and scopes the `key` requirement to `env_vars`.
3. **"Some results carry `_actions`…"**, and the read-only `_actions` leak is filed as #390.
4. **`TESTED_RANGE` moved to `src/lib/tested-range.ts`**; doctor and the instructions both import it, and `index.ts`'s lazy load of doctor stays intact.
5. **Stale-tool-name guard** added as suggested, against `TOOL_ANNOTATIONS`, with `action`/`instance`/`key` allowed as arguments.
6. **"Every tool except `list_instances`"**, and the read-only and HTTP variants are jest-snapshotted alongside the single and fleet texts (evals has no harness for those two modes).
7. Nits: harness comment back on `toolInfo`; the module comment now says the text is a function of the server's own configuration (including the validated default instance name); "a 405, or a 404 on an older instance".

844 → 848 tests locally, lint clean, evals snapshots regenerated.

Issue #340, opened 2026-08-06
OAuth hardening for HTTP mode: CIMD over DCR, 401 shape, PRM paths, SSRF guards


StuMason:
Auth findings from a review of the current spec and Anthropic's connector docs, relevant to the OAuth AS built in #303.

## 1. DCR is deprecated — advertise CIMD

`2026-07-28` deprecates OAuth Dynamic Client Registration (RFC 7591) in favour of **Client ID Metadata Documents**. Anthropic explicitly recommends CIMD over DCR at directory scale, because DCR registers a brand-new client on every fresh connection — which for us also means unbounded growth in the OAuth state file.

Claude selects CIMD only when the AS metadata carries **both**:

- `client_id_metadata_document_supported: true`
- `"none"` in `token_endpoint_auth_methods_supported`

We already advertise `"none"`, so this is mostly adding the first flag plus the document-fetch path. Keep `/register` as fallback for older clients.

## 2. Confirm the 401 shape

The most common connector failure is returning **200 with `isError: true`** instead of a transport-level 401. A 200 produces no auth prompt at all — the client never runs OAuth and never retries. It has to be:

```
HTTP/1.1 401 Unauthorized
WWW-Authenticate: Bearer resource_metadata="https://…/.well-known/oauth-protected-resource", scope="…"
```

and it has to be gated at the HTTP layer **before** the JSON-RPC body reaches the SDK. Worth an explicit integration test rather than an assumption.

## 3. Protected-resource metadata at both paths

RFC 9728 metadata should be served at **both** `/.well-known/oauth-protected-resource` and the path-suffixed variant, and the `resource` value must exactly match the URL the user typed. A trailing-slash or scheme mismatch here fails silently.

## 4. Redirect URIs to register

- `https://claude.ai/api/mcp/auth_callback`
- Port-agnostic `http://localhost/callback` and `http://127.0.0.1/callback` for Claude Code

## 5. Timeouts

Claude gives **10 s** for discovery, registration and token exchange, **30 s** for refresh. Anything slower fails the connection rather than degrading.

## 6. Security items now normative in the spec

- **Token passthrough forbidden** — never accept a token not issued for us; validate audience per RFC 8707 (we do this).
- **RFC 9207 `iss`** validation for mix-up attacks.
[...]

PR #389, opened 2026-09-09
feat(http): Client ID Metadata Documents, DCR kept as fallback (#340)

StuMason:
Item 1 of #340. Not closing the issue: the redirect-URI allowlist, RFC 9207 validation on the client side and the 10 s discovery measurement stay open there.

**What changes.** The authorization server advertises `client_id_metadata_document_supported: true` (alongside the `none` auth method it already advertised, which is the pair Claude keys on) and accepts an https URL as `client_id`. The registration is fetched from that URL and validated per the draft:

- identifier URL: https, no userinfo, a path component, no `.`/`..` segments (checked on the string as sent, since `new URL()` collapses them), no fragment
- document `client_id` equals the identifier by simple string comparison, no normalisation on either side
- `redirect_uris` under exactly the rules `/register` applies (https or http loopback, no fragment), and the authorize leg still requires an exact match
- `token_endpoint_auth_method` absent or `none`; any shared-secret method or a `client_secret` in the document is rejected
- the consent page shows `Name (host)` so the person sees where the registration came from

**Fetch path.** `fetchPublicJson` from `src/lib/ssrf.ts`, as the issue's own note requires: public addresses only, pinned DNS, redirects refused, 8 KB and 10 s caps. Documents are cached in memory for an hour and never written to the state file; fetch and validation errors are never cached.

**Why this shape.** `resolveClient(client_id)` is an async pre-step run on the three legs that carry a client_id (authorize GET, authorize POST, token). The existing synchronous lookups stay synchronous and consult one `lookupClient` that sees both registered and document clients, so the 16 existing test call sites and the DCR flow are untouched. `/register` is unchanged and remains the path for clients that predate the document.

**What it fixes beyond the spec.** DCR registers a new client on every fresh connection, so the `/data` volume grew by one client per connection forever. Document clients never touch the volume.

**Tests.** 21 new cases: metadata flag, full code flow on a URL client with one fetch serving both legs, no-op for registered and non-URL ids, six identifier-URL rejections that never fetch, six document rejections that are not cached, fetch failure surfaced as `invalid_client` and retried, TTL expiry re-fetching so a client can change its redirect URIs, exact redirect match, state file never containing the document client, and the real guard refusing loopback and the metadata-service address through all three HTTP legs. 844 total, lint clean.


StuMason:

1. **GET leg limited and blinded.** `GET /authorize` with a URL `client_id` now draws on `authLimiter` under a `cimd:` bucket (registered ids keep their unlimited in-memory page). A fetch failure renders one generic sentence; the reason (did not resolve / 302 / timeout / too large) goes to `console.error` with the host. Concurrent resolves for the same `client_id` share one in-flight fetch.
2. **Eviction.** `prune()` now sweeps document clients past TTL plus grace, and the map is capped at 512 with oldest-first eviction (entries re-inserted on refresh so Map order stays by age). A test drives 600 distinct ids and then advances the clock through a prune.
3. **One code.** `validateRedirectUris` failures inside a document are re-thrown as `invalid_client` ("lists an unusable redirect_uri: …"); the table test now asserts `code: 'invalid_client', status: 401` for every document rejection.
4. **Grace on refresh.** A known client whose document host stops answering keeps its last good document for a day measured from that fetch, then gets `invalid_client` until the host is back. `docs/http-mode.md` says so in the connect section.

Smaller ones: `resolveClient` refreshes within five seconds of expiry so the synchronous lookup that follows never sees a lapsed entry; the comment now matches the 8 KB constant; each raw path segment is percent-decoded before the dot check; the TTL tests run on fake timers; `docs/security.md` names the one attacker-chosen outbound fetch and what guards it.

848 tests locally, lint clean. The codecov patch miss was the default-fetcher and error branches; the new tests cover the limiter, the log line, the grace path and eviction.

Issue #384, opened 2026-09-09
Doctor: MCP tool front door, error suffix, masking evals, fleet coverage


StuMason:
What is left of #368 now the startup self-check and the doctor CLI have shipped in 3.1.0.

1. **`doctor` as an MCP tool**, same checks as the CLI. Schema change, so it comes with the contract snapshot and eval churn; own PR.
2. **Every tool error that matches a doctor check ends with "run `npx @masonator/coolify-mcp doctor`"**, so the user who never reads a README still finds it.
3. **Secret-masking evals over doctor output**, asserting no token or header value ever appears in any check line or the `--json` shape.
4. **Doctor over the fleet registry**: run the checks per instance and report them keyed by name, including the `COOLIFY_INSTANCES` per-entry shape checks and a deadline on `list_instances`.
5. **`StartupCheckResult` carries `{variable, message}`** so doctor's gates match on the variable name, not on message prose.

Each is its own PR.

StuMason:
From the #391 review: `DocsSearchEngine.status()` reports whether `search_docs` is serving the bundled or the live index and how old the bundle is. Worth a doctor line ("docs index: bundled, 40 days old; live refresh failed") once the MCP tool front door lands.

PR #391, opened 2026-09-09
feat: bundle the Coolify docs index so search_docs works offline (#372)

### LINKED ISSUE #372: 3.1: bundle the docs-search index, stop depending on coolify.io at runtime
## Why

`search_docs` fetches `https://coolify.io/docs/llms.txt` at call time (`src/lib/docs-search.ts`). That means docs search fails offline, fails behind egress-restricted networks, fails when coolify.io hiccups, and adds a network round-trip to every cold search. A competing server (frndchagas/coolify-mcp) bundles a local index and takes no network dependency at all — on this one axis their behaviour is strictly better, and there's no reason to leave it that way.

## Proposal

1. Fetch and build the MiniSearch index at **build/release time**; ship it in the package (it's an index over llms.txt, not the 40MB full dump — measure, but expected small).
2. Runtime behaviour: serve from the bundled index immediately; optionally refresh from the live URL in the background with a short timeout, falling back silently to the bundle. Never block a search on the network.
3. A release-time freshness check: the publish workflow warns when the bundled index is older than N days against the live one, so the bundle tracks upstream docs without manual thought.
4. Result URLs keep pointing at live coolify.io pages (deep-reading stays online-only, as today — that's the right split).

## Acceptance

- `search_docs` returns results with networking disabled (eval fixture runs with a refused fetch).
- Package size delta reported in the PR; bundle rebuilt by the publish workflow, not committed by hand.
- Behaviour unchanged when online except faster.

StuMason:
Closes #372.

`search_docs` fetched `https://coolify.io/docs/llms.txt` on first call, so it failed offline, behind egress rules, and whenever coolify.io hiccupped. The index now ships inside the package.

**Runtime.** `src/data/coolify-docs.json` (the raw `llms.txt`, 271 pages, 46 KB, with `fetched_at`, ETag and Last-Modified) is copied to `dist/data` by the build. The first search parses and indexes it synchronously and starts one background refresh from the live URL with a five-second timeout; if that returns a file that parses, the live entries replace the bundled ones for the rest of the process. A search never waits on the network and never fails because of it. `status()` reports which source is being served. The one deliberate noise: a live file that parses to zero entries is logged once to stderr, because that is an upstream format change the next refresh would otherwise carry into the bundle unnoticed.

**Keeping it fresh.** `scripts/build-docs-index.mjs` fetches the live file, refuses to overwrite the bundle if it parses to fewer than 100 pages, and writes only when the content changed (so a no-op refresh produces no diff). `npm run docs:index` refreshes, `npm run docs:index:check` reports drift. The publish workflow runs the refresh before `npm run build` with `continue-on-error`, so a release carries the docs as they stand that day and a coolify.io outage ships the committed copy instead of blocking the release. A new weekly workflow opens or updates a PR when the committed bundle falls behind; nothing merges on its own.

**One divergence from the issue as written.** It asked for the bundle to be built by the publish workflow and not committed. The bundle is committed as well, generated by the script, because the tests and the container build need it without network and a package whose contents are not in git is harder to reason about. The publish-time refresh still happens as asked.

**Tests.** The engine suite is rewritten for the new contract: serves from the bundle while a refresh hangs forever; exactly one refresh per process, swapped in when it parses; no headers off-estate (the #373 check, kept); bundle keeps serving through a network error, a 404 and a zero-entry live file; the zero-entry case logs once; a bundle that parses to nothing throws as a broken build; and the acceptance case from the issue, the real shipped bundle answering "docker compose" with fetch rejecting outright. 840 tests total, lint clean, `npm pack` confirms `dist/data/coolify-docs.json` is in the tarball.


StuMason:

1. **Canary restored.** `docs-search.integration.test.ts` now fetches `llms.txt` itself and asserts the runtime parser reads more than 100 entries with zero unparsed link lines; the second case searches from the bundle and then waits for the background swap. Ran it here against the live file: green.
2. **One parser.** The parser moved to `src/lib/docs-index-parse.mjs` (plain JS, consumed through `allowJs`) and both the engine and the script import it. `validateIndex` refuses a file with any link-looking line the parser rejects, checked before the entry count so the reason is the specific one; the "— description" reshape you described is now a test that fails the script.
3. **Lazy bundle.** Nothing is read in the constructor; a test asserts the loader is untouched until first use.
4. **Heal path.** A missing or zero-entry bundle tries the live index once, in the foreground; if that fails too, every search throws naming `npm run docs:index`, and the attempt is not repeated per search.
5. **Workflow.** `git fetch origin "$branch" || true` before the lease push; any exit other than 0 or 2 fails the step; the PR body says CI needs a push and that the file was already validated with the runtime parser.
6. **Timeout covers the body**; the abort is disarmed in `finally` after `text()`.
7. **ETag wired.** `If-None-Match` rides the refresh when the bundle has an ETag and a 304 means "current"; the #373 assertion now checks the exact header set instead of none.
8. **Script tests** under `scripts/__tests__/build-docs-index.test.ts`: validation, no-op write keeps `fetched_at`, `--check` exit codes and age line, and untouched-on-failure.
9. `status()` into doctor is left as a follow-up; noted on #384 with the other doctor slices.

839 tests, lint clean, `--check` reports the bundle matches live.

PR #392, opened 2026-09-09
feat: ship MCP prompts and resources, not just tools (#371)

### LINKED ISSUE #371: 3.1: ship MCP prompts and resources, not just tools
## Why

Coolify's built-in MCP ships 2 MCP resources (`coolify://overview`, `coolify://application/{uuid}`) and 2 prompts (`troubleshoot_application`, `explain_failed_deploy`) alongside its tools. We ship 45 tools and nothing else. Prompts and resources are cheap, clients surface them in their UI (slash-command style), and they're the difference between "a pile of tools" and "a product that knows its own workflows". We already have the workflow knowledge — `diagnose_app` and `find_issues` encode it — this exposes it at the protocol level.

## Proposal

1. Prompts (thin orchestration over existing tools, no new API surface):
   - `troubleshoot_application` — name/uuid in, drives diagnose_app + recent deployment logs + env sanity into a guided diagnosis.
   - `explain_failed_deploy` — deployment uuid in, walks the log tail (through `asUntrustedLogs`, unchanged) with the known failure patterns.
   - `estate_health` — find_issues + infrastructure overview into a morning-coffee summary.
2. Resources:
   - `coolify://overview` — the infrastructure overview snapshot.
   - `coolify://application/{uuid}` — the sanitized application detail.
   Resource content goes through the same central sanitizer as tool responses — a resource read must never be a masking bypass.
3. Contract snapshots extend to prompts/resources so the eval gate covers them.

## Acceptance

- Prompts and resources listed over both stdio and HTTP mode.
- Masking eval: `coolify://application/{uuid}` for an app with webhook secrets shows the same masked shape as `get_application`.
- tools/list token budget unchanged (prompts/resources are separate lists).

StuMason:
Closes #371.

The server has exposed one of MCP's three primitives. This adds the other two: three prompts a client shows as slash commands, and two resources a client can attach. No new Coolify endpoint, no new dependency — `registerPrompt` and `ResourceTemplate` are already in SDK v2.

## Prompts

| Prompt | Argument | Workflow |
| --- | --- | --- |
| `troubleshoot_application` | `query` (name, UUID or domain) | `diagnose_app`, then `logs`, then env sanity, ending in a cause and the smallest fix |
| `explain_failed_deploy` | `deployment_uuid` | build log through `deployment` get, which stage broke, quoting the evidence |
| `estate_health` | none | `find_issues` plus the overview, worst first |

Fleet mode adds the same optional `instance` the tools take.

### They make no API call

`prompts/get` is synchronous, has no elicitation channel and no error affordance, so a prompt that pre-fetched would turn a slash command into a multi-second stall that can fail with something the human cannot act on.

The bigger reason is security. Embedding container or build output in the returned message would place attacker-influenceable text in a **user-role** message, outside the `asUntrustedLogs` boundary the tool layer puts around exactly that text (`evals/FINDINGS.md` #4). Naming the tool and letting the model call it keeps the log on the one path that frames it as data. The builders are pure string functions and a test spies on the client to prove it.

### A prompt whose tools are missing is not listed

This was the design's one real wall. Read-only mode (#303) registers no mutating tool, and consolidation puts two pure reads under destructive tools: `env_vars` list and `deployment` get. So "which tools exist" is a per-mode fact, and a prompt naming `env_vars` names a tool a read-only server does not have.

Resolved structurally rather than with a comment nobody maintains. `defineTool` records what it registered; `definePrompt` takes `requires` and skips the whole prompt when a required tool is absent; the builders take `ctx.has()` and drop an individual step.

Concretely, on a read-only server:

- `explain_failed_deploy` is **not listed at all** — the build log is reachable only through `deployment`, so the workflow cannot run, and a dead-end slash command is worse than none.
- `troubleshoot_application` is listed with its env step dropped.
- `estate_health` is unchanged.

## Resources

`coolify://overview` and `coolify://application/{uuid}`, instance-scoped in fleet mode (`coolify://staging/overview`), because reading production's overview while believing it is staging's is the mistake a second instance invents. Every application is listed as a concrete entry so a human can pick one instead of knowing a UUID; an unreachable instance drops out of the listing rather than blanking it.

**Masking.** Reads go through the same client as every tool call, so `deepSanitize` applies. The contract suite asserts the resource payload equals `get_application`'s byte for byte rather than assuming it, with a planted fixture secret so the paired assertion can actually fail. No URI takes a `reveal`: `get_application` has one because a caller justifies it in the moment, whereas a resource URI is a durable handle a client may cache, re-read or paste.

`get_infrastructure_overview` moved into a shared private method, so the tool and the resource cannot drift into two answers.

## Contract

`tools/list` is untouched — these are separate lists over separate methods, which is what #371 asked for. The evals harness now lists all three surfaces from one boot, with `_prompts.json` / `_resources.json` snapshots and fleet variants, their own budgets, and a check that no prompt names a tool the server did not register.

## Verification

- `npm test` — 844 passing, 15 suites. `prompts.ts` at 100% statements.
- `npm run snapshots` — 23 contract tests passing, single and fleet.
- `npm run lint`, `npx tsc --noEmit`, evals `typecheck` + `lint` clean. No new warnings.
- `site` build and tests pass; the new doc lands in `/llms.txt` and `/llms-full.txt`.

## Notes

- Client support is uneven. Prompts surface as slash commands in Claude Code and Claude Desktop; resources appear as attachments in Desktop and by mention in Claude Code. A client supporting neither loses nothing.
- Instance-scoped resource URIs are shipped here rather than deferred, since the alternative was a fleet reading the default instance under an unscoped URI.
- Like the other open 3.2 branches this adds a line at the same `### Added` anchor in CHANGELOG.md, so expect a conflict there on the second and later merges.


StuMason:
Review round applied in 83a69a8b. Every finding actioned.

**Two were real bugs my own snapshots had encoded:**

1. Listed resources inherited the template `title`, so every application rendered as the identical row "Application detail". Each entry now carries its own title, with a test asserting no two listed resources share one.
2. An empty `{instance}` capture fell through to the default instance, since `registry.get('')` returns the default exactly as `get(undefined)` does. Now throws, as does an empty `{uuid}` (which would have reached Laravel's index route and returned every application under a URI claiming one).

**The rest:** the `instructions.ts` reference was to an unmerged sibling branch and is reworded; `registeredPrompts` now backs a boot-time roster check instead of being write-only; the "names only real tools" check is inverted to fail closed with a `NOT_A_TOOL` list for argument names; coverage added for fleet application reads, the `estate_health` plural branch, a rejecting read and a quoted application name; prompt arguments are `JSON.stringify`d; `{instance}` gains completion on both templates; the user-role assumption about prompt arguments is now stated.

**Not taken:** the `resources/list` fan-out cache. A short TTL helps the second listing onwards but not the first, and it buys staleness on a surface whose job is to be current. The timeout half may be the more valuable one. Filed as #393 with the reasoning, and the cost is named in the code.

The rejecting-read assertion sits in the unit tests rather than the contract suite: the eval fixture answers an unknown uuid with HTTP 200 and a not-found body where real Coolify sends a 404. Worth knowing separately.

The contract gate failure was mine and is fixed: `evals/` has its own prettier config, so those files must be formatted from inside `evals/`.

The three `build (N.x)` failures are not from this branch. They are GHSA-7w5x-hrqm-74c2 in `smol-toml`, which fails on all five open PRs. Fixed separately in #394, which is fully green. Merge that first and this goes green on rebase.

Now: 847 jest tests, 26 contract tests, tsc / lint / evals typecheck clean.

Issue #417, opened 2026-09-14
HTTP mode outside the image: default MCP_OAUTH_STATE_FILE crashes the process on first registration and on SIGTERM


artgas1:
## What happens

Run HTTP mode directly on a machine (not in the image) without setting `MCP_OAUTH_STATE_FILE`. It starts, `/healthz` answers, and the first thing that persists OAuth state kills the process:

```text
coolify-mcp http mode on :13473 (public: http://127.0.0.1:13473, read-only)
Error: ENOENT: no such file or directory, mkdir '/data'
    at mkdirSync (node:fs)
    at OAuthProvider.writeState (dist/lib/oauth.js)
    at Timeout._onTimeout (dist/lib/oauth.js)
```

The default `/data/oauth-state.json` only exists in the image (the Dockerfile creates `/data`). Outside it, `writeState()` throws from the debounced `persist()` timer, which is an uncaught exception, so the process exits 1.

Two ways to hit it:

1. **First client registration.** `POST /register` returns `201` with a `client_id`, then about 250 ms later the debounced write throws and the server is gone. The client has been told it registered against a server that no longer exists.
2. **Plain shutdown.** Start it, send `SIGTERM` with nothing written yet. `shutdown()` calls `flush()`, `writeState()` throws the same `ENOENT`, exit code `1` instead of `0`.

## Repro

`main` at 19771df5 (package.json 3.4.0), Node 24.16.0, macOS. Placeholder Coolify values are enough, nothing reaches Coolify:

```bash
npm run build
COOLIFY_BASE_URL=https://coolify.example.invalid \
COOLIFY_ACCESS_TOKEN=placeholder \
MCP_PUBLIC_URL=http://127.0.0.1:13473 MCP_ALLOW_INSECURE_HTTP=true \
MCP_PORT=13473 MCP_READONLY=true \
node dist/http.js &

curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:13473/healthz      # 200
curl -s -o /dev/null -w '%{http_code}\n' -X POST -H 'content-type: application/json' \
  -d '{"redirect_uris":["http://127.0.0.1/callback"],"client_name":"repro","token_endpoint_auth_method":"none"}' \
  http://127.0.0.1:13473/register                                             # 201
sleep 2; curl -s -o /dev/null -w '%{http_code}\n' http://127.0.0.1:13473/healthz  # 000, process exited 1
```

[...]

PR #418, opened 2026-09-14
feat(http): MCP_HOST sets the listen address

artgas1:
## Summary

HTTP mode calls `server.listen(port)` with no host, so it always binds every interface. That is right in a container, where the platform proxy reaches it over the container network. Run `coolify-mcp-http` on a workstation and the same default puts a process holding a Coolify API token on every network the machine is attached to: `lsof` shows `*:PORT`, and `/healthz` answers on the machine's LAN address. `/mcp` itself still needs OAuth, but the authorize page, `/register` and the health endpoint answer anyone who can reach the port, and there is no setting to prevent it.

This adds `MCP_HOST`. Unset or blank keeps today's behaviour, so existing deployments are untouched.

## Changes

- `listenOptionsFromEnv(env)` in `src/lib/http-server.ts` resolves `{ port, host? }`. Port precedence is unchanged (`MCP_PORT`, then `PORT`, then `8080`). `MCP_HOST` is trimmed, and when it is unset or blank the `host` key is omitted rather than set to `undefined`.
- `src/http.ts` passes that object to `server.listen()`. The startup line keeps its `on :8080` form when no host is set, which the docs tell people to look for, and names the address when one is (`on 127.0.0.1:8080`, `on [::1]:8080`).
- `MCP_HOST` joins the HTTP variables checked by `checkStartupConfig`, so an unexpanded `${VAR}` is a readable startup error instead of an `ENOTFOUND` crash from `listen()`. stdio ignores it, like `MCP_PUBLIC_URL`.
- `docs/http-mode.md`: a row in the configuration reference and a short "Running outside a container" note. `CHANGELOG.md`: an Unreleased entry.

## Test plan

- [x] Tests added. `listenOptionsFromEnv`: unset and blank omit the host, port precedence, trimming, and two real-socket cases (`MCP_HOST=127.0.0.1` binds `127.0.0.1`; unset binds `::` or `0.0.0.0`). `describeListen` for the log line. `startup-check` covers `MCP_HOST` in the placeholder check and on stdio. Replacing `return host ? { port, host } : { port }` with `return { port }` and removing `MCP_HOST` from the checked list turns 4 of these red.
- [x] The CI commands locally (Node 24.16.0): `npm audit --audit-level=high`, `prettier --check .`, `npm run lint` (0 errors, no new warnings in touched files), `check:spec-drift`, `check:chunk-drift`, `check:tool-count`, `npm run build`, `npm run test:coverage` (20 suites, 1041 tests). One run hit a 5 s timeout in `mrtr-confirmation.test.ts` on a heavily loaded machine; that file passes alone and the full suite passed on rerun.
- [x] Manually, built `dist/http.js` with placeholder Coolify values, `MCP_READONLY=true` and a temp `MCP_OAUTH_STATE_FILE`:

| Setting | `lsof` listener | `/healthz` on 127.0.0.1 | `/healthz` on LAN address | Startup line |
| --- | --- | --- | --- | --- |
| `main`, unset | `*:13472` | 200 | 200 | `on :13472` |
| this branch, unset | `*:13472` | 200 | 200 | `on :13472` |
| this branch, `MCP_HOST=` | `*:13472` | 200 | 200 | `on :13472` |
| this branch, `MCP_HOST=127.0.0.1` | `127.0.0.1:13471` | 200 | connection refused | `on 127.0.0.1:13471` |
| this branch, `MCP_HOST=::1` | `[::1]:13472` | refused (`[::1]` answers 200) | connection refused | `on [::1]:13472` |

Unrelated, found while testing: without `MCP_OAUTH_STATE_FILE`, running outside the image dies on the first registration and exits 1 on `SIGTERM`, because `/data` does not exist. Filed as #417, not touched here.

https://claude.ai/code/session_01NPiG3ikPXhAfVoBso27Nv8


StuMason (approved):
Thanks, this is a clean one. Unset keeps the exact old `listen()` call, and the real-socket tests are a nice touch. One possible follow-up: a `server.on('error')` handler so a mistyped `MCP_HOST` names itself instead of dumping an ENOTFOUND stack. Happy to merge as is.

PR #424, opened 2026-09-16
fix: HTTP mode outside the container no longer dies on its first state write (#417)

### LINKED ISSUE #417: HTTP mode outside the image: default MCP_OAUTH_STATE_FILE crashes the process on first registration and on SIGTERM

StuMason:
Fixes #417.

## What was wrong

The OAuth state file defaults to `/data/oauth-state.json`, a directory that only exists in the image. Run `dist/http.js` on a workstation without `MCP_OAUTH_STATE_FILE` and the server started, answered `/healthz`, accepted the first `POST /register` with a 201, then exited 1 about 250ms later when the debounced state write threw from a timer. A plain SIGTERM did the same through the shutdown flush. The code dates from HTTP mode's first release in 3.0.0, and the docs section #418 added for running outside a container never mentioned the state file.

## What changes

- **Boot check.** `ensureStateFileWritable()` in `startup-check.ts` resolves the state file's directory, creates it 0700 if missing (the write path has always done that) and writes and removes the temp file the provider uses. A real write rather than an access check, so a stale `.tmp` left by another user is caught too. `http.ts` lists any failure with the other "cannot start" problems, naming the path and the errno. The message explains where the default comes from only when nothing was configured. A recursive mkdir on an existing directory is a silent no-op even when the process cannot write into it, which is the root-owned `/data` case, so the write after it is the step that catches the report.
- **Runtime write.** `writeState()` catches and latches: one line naming the path and error on the way into broken, one when a write succeeds again, and `/healthz` reports `"persistence": "degraded"` in between. The server keeps serving from memory. Nothing in use is lost, only survival across the next restart.
- **Docs.** A paragraph in "Running outside a container", a note in the env table, a sentence in PRIVACY.md, and a changelog entry under Unreleased with upgrading notes, including the bind-mount case that used to start and die and now refuses to start.

Container deployments are unaffected: the image's `/data` volume passes the new check.

## Tests

Unit tests in `startup-check.test.ts` cover the in-memory setting, a writable directory, a missing directory being created, nothing left behind on success, a stale temp file an access check would pass, the mkdir failure with both message variants, and the exists-but-unwritable case (skipped as root, where it is unobservable). `oauth.test.ts` covers the shutdown flush logging instead of throwing with tokens still verifying afterwards, the debounced timer write with real timers plus the latch on the next failure, recovery clearing the flag with its own line, and the degraded field on `/healthz`. That last test fails on the old code with the uncaught exception from the report.

End to end against the built `dist/http.js` as a non-root user, placeholder Coolify values, `MCP_HOST=127.0.0.1`:

| Run | Setup | Result |
| --- | --- | --- |
| A | `MCP_OAUTH_STATE_FILE` unset | Refuses at boot, exit 1, never listens. stderr names `/data/oauth-state.json`, `EACCES: permission denied, mkdir '/data'` and the fix. |
| B | Writable nested path that does not exist yet | Directory created, register 201, healthz 200 a second later, SIGTERM exits 0, state file written 0600. |
| C | Writable path, then `chmod 0500` on the directory while running, then `chmod 0700` back | Register 201, healthz still 200 with `"persistence": "degraded"`, one `serving from memory` line, a second register adds no line, register after the chmod back writes a `state persists again` line and healthz drops the field, SIGTERM exits 0, directory is 0700. |

Before this change, run A was the reporter's repro: 201 then exit 1.


StuMason:
Review actioned in the second commit.

**Taken**
- (1) Real write probe: the check writes and removes the temp file the provider uses. A directory in the temp file's place (EISDIR, works as root too) is the test for the stale-`.tmp` case an access check would pass.
- (2) Latch: logs on the transition into broken and on the first write that succeeds again. The "exactly once" assertion in the timer test now pins a guarantee, and the same test flushes a second failure to show it says nothing new. A recovery test removes the blocker and checks the flag clears and the file lands.
- (3) `/healthz` adds `"persistence": "degraded"` while latched. Existing shape unchanged otherwise. Tested, and run C in the description shows it end to end, including the field clearing after recovery.
- (4) Upgrading note for the daemon-created bind mount, and a line saying the failure is loud at boot now rather than after the first registration.
- (5) Renamed to `ensureStateFileWritable` with the side effect documented on the export. `mkdir` is 0700 in both places. `DEFAULT_OAUTH_STATE_FILE` is one exported constant. Empty-path test renamed to say what it is. PRIVACY.md has the sentence.

**Left**
- The directory is still created when boot is doomed for another reason. Skipping the probe in that case would drop the state-file problem from a multi-problem boot, and one boot listing everything is the rule the check exists for.
- `behindAFile` stays duplicated across the two test files. Small enough that a shared helper would be more file than function.
- The `http.ts` argument wiring stays untested, consistent with the rest of that file. Run A covers the default path for real.

